PR #339, opened 2022-05-31
Pull Spring and Fall terms out of the Term select - IN PROCESS

bryantal:
- Spring and fall will no longer be displayed in the term selectpicker
- added checkbox for spring and fall
- made sure only one set of spring/fall only appears for each AY
- Release/rehire modal now shows the correct term

To test:

1. Have AY as well as either spring or fall term open for a specific school year
2. Select the AY and then check the fall or spring only term


BrianRamsay:
They switched back to Fall and Spring terms 🤦🏼

PR #340, opened 2022-05-31
Fixes My Current Students page to go to right place

bryantal:
- Fixes the “My Students” tab to display the correct information
- Removes a dead javascript import for jspdf (there are two imports)
- Fixed a small formatting issue on “No longer a student” display (extra line break)


tylerpar99:
The issue looks old, and can probably be closed. @BrianRamsay @Andersonstettner

PR #341, opened 2022-05-31
Add dates to primary checks

bryantal:
This PR includes:
1. Refactor of supervisor portal tabs to include “Pending”, “Past”, and any active terms based on the end dates of the term. Leverages Datatables to do this filtering (instead of lots of backend code).
2. Refactor of supervisor portal to include all LSFs (instead of all students). You should see duplicates if a student has more than one LSF given the filter in item 1.
3. Refactor of supervisor portal’s organization of information.
4. Fixes the My students menu item to the correct page.
5. Hotfix of the Labor History buttons for students who are no longer at the college. Should display “Show evaluations” if one exists, otherwise, it should show no buttons and the “No buttons available” warning message.
6. Fixes the Download button to include all LSFs that are in the Datatable (instead of what the backend says to send, which was not always accurate based on the user’s current view). The Downloader pulls across a paginated table, also.


BrianRamsay:
Too old, page has been overhauled since.

Issue #358, opened 2022-08-18
Fix Note management for third party approvers


BrianRamsay:
The financial aid and saas reviewers should be able to manage notes on their approval pages.

Currently there is a disabled input box that shows only labor notes. We should change it so that they have a way to view and add notes without having to approve and deny the form. Same interface pretty much as the current View Notes action on the overloads page.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix Note management for third party approvers #358](https://trello.com/c/MOImVhqE/485-fix-note-management-for-third-party-approvers-358)

Andersonstettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix Note management for third party approvers #358](https://trello.com/c/V3PqzJ21/484-fix-note-management-for-third-party-approvers-358)

PR #367, opened 2022-09-05
Add note section for third party Approvers 

sreynit02:
This PR fixes issue #358 : "The financial aid and saas reviewers should be able to manage notes on their approval pages. Currently there is a disabled input box that shows only labor notes. We should change it so that they have a way to view and add notes without having to approve and deny the form. Same interface pretty much as the current View Notes action on the overloads page."

Did: Added a textarea and a Save Note button on `/admin/saasOverloadApproval/formId`. 
Added a function called `overloadNoteInsert` in `financialAidOverload.js `file to handle note insertion from the financialOverload page.

Test: 
- Submit an Overload form for a student. An Overload Form is created when the total labor hour for student is over 20. 
- Log in as Financial Aid and navigate `http://0.0.0.0:8080/admin/financialAidOverloadApproval/<formId>`. Under "Financial Aid Note", add a note and click "Save Note". The page should reload and clear the text. Check the note table in php or mysql workbench, a note entry should be created for the note you just input.
-  Log in as Saaa admin and navigate `http://0.0.0.0:8080/admin/saasOverloadApproval/<formId>`. Under "Saas Note", add a note and click "Save Note". The page should reload and clear the text. Check the note table in php or mysql workbench, a note entry should be created for the note you just input.

### DIFF
--- a/app/static/js/financialAidOverload.js
+++ b/app/static/js/financialAidOverload.js
@@ -76,15 +76,3 @@ function insertOverloadNote(textareaID, buttonID) {
       }
   });
 }
-
-export function setNotesLimit(textboxId, labelId){
-  var maxCharacters = 250;
-  var textLength = 0;
-  var textVal = $("#" + textboxId).val();
-  var textLength = textVal.length;
-  $(labelId).text("Remaining Characters: " + (maxCharacters - textLength));
-  if (textLength > maxCharacters){
-    $("#" + textboxId).val(textVal.substring(0, maxCharacters));
-    $(labelId).text("Remaining Characters: " + 0);
-  }
-}
--- a/app/static/js/financialAidOverload.js
+++ b/app/static/js/financialAidOverload.js
@@ -77,13 +77,12 @@ function insertOverloadNote(textareaID, buttonID) {
   });
 }
 
-function setNotesLimit(textboxId, labelId){
+export function setNotesLimit(textboxId, labelId){
   var maxCharacters = 250;
   var textLength = 0;
   var textVal = $("#" + textboxId).val();
   var textLength = textVal.length;
   $(labelId).text("Remaining Characters: " + (maxCharacters - textLength));
-  console.log(maxCharacters - textLength);
   if (textLength > maxCharacters){
     $("#" + textboxId).val(textVal.substring(0, maxCharacters));
     $(labelId).text("Remaining Characters: " + 0);
--- a/app/static/js/financialAidOverload.js
+++ b/app/static/js/financialAidOverload.js
@@ -58,9 +58,10 @@ function openApproveDenyModal(status){
     $("#finOverloadModal").modal("show");
   }
 }
+
 function overloadNoteInsert(textareaID, buttonID) {
-  var laborNotes = $("#" + textareaID).val(); //this is getting the id of the labor notes text area
-  var data = JSON.stringify(laborNotes);
+  var overloadNotes = $("#" + textareaID).val();
+  var data = JSON.stringify(overloadNotes);
   $("#" + buttonID).on('submit', function(e) {
     e.preventDefault();
   });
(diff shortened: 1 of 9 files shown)


app/static/js/financialAidOverload.js
```diff
@@ -58,3 +58,21 @@ function openApproveDenyModal(status){
     $("#finOverloadModal").modal("show");
   }
 }
+
+function overloadNoteInsert(textareaID, buttonID) {
```
tylerpar99: A good variable name, ideally verbs come first. I would recommend the variable be switched to insertOverloadNote.

app/static/js/financialAidOverload.js
```diff
@@ -58,3 +58,21 @@ function openApproveDenyModal(status){
...
+  var overloadNotes = $("#" + textareaID).val();
```
tylerpar99: Is this a single note? or a list of notes? I would change it to overloadNote if it is a single note being grabbed. However, if it is multiple you can ignore this comment.

tylerpar99 (changes requested):

bledsoef (dismissed):

tylerpar99 (approved):

BrianRamsay (changes requested):
The financial aid notes should be added as financial aid notes, not 'Labor' notes. It looks like this is broken for the notes added from the "View Notes" action on the overload forms list as well (broken before you).

Also, there is no way to view the notes on the approval screen itself, which was part of the ticket. It can be the same UI as the View Notes action has, which will make it simpler, and you can replace the Labor Notes text area.

sreynit02:
@BrianRamsay did you mean this "View Note" on this modal?

image

AndersonStettner (commented):
Add Label to notes box for SAAS

sreynit02:
> 

fixed

AndersonStettner (changes requested):
Currently, the only way to see notes that have been left is to back out of the overload approval form and view the notes modal but we also need to see all of the notes that have been left while looking at the overload approval form. It would also need to be updated if a user leaves a new note on the page.

AndersonStettner (changes requested):
If the user adds a long note, after a certain point, the note will not actually be submitted. The page will be reloaded and the note will disappear from the input box but it will not show up in the Notes log or in the database.

Karina-Agliullova:
image
Looks great! Remove the labels for other note modals. Having the title for notes is enough. If we include the label, it is kind of redundant

gahimbaref:
There is no limit to how much you can expand the text box
image

AndersonStettner:
> There is no limit to how much you can expand the text box image

I would say this is not actually an issue. Being able to expand the box is an accessibility thing so we would want to allow it to get as big as the user wants, regardless of the size of the modal.

sreynit02:
> If the user adds a long note, after a certain point, the note will not actually be submitted. The page will be reloaded and the note will disappear from the input box but it will not show up in the Notes log or in the database.

Fixed!

gahimbaref:
It seems as though the number of notes is not being incremented correctly. For example, there might be 4 notes, but it will show up as "View Notes(2)". 

Also, if you make heggens a saasAdmin but not a labor admin and not a financial aid admin, when you add a note, the number stays the same.

image

AndersonStettner (changes requested):
It's looking good. Two things, you didn't remove the second notes label in the view modal and, the sorting of the notes is different, they should be sorted the same way wherever they are.

sreynit02:
> It's looking good. Two things, you didn't remove the second notes label in the view modal and, the sorting of the notes is different, they should be sorted the same way wherever they are.

fixed!

gahimbaref (approved):
looks good!

BrianRamsay (approved):
Looks good! Thanks.

Issue #366, opened 2022-09-02
Faculty and staff should be able to search their departments form's and students


bledsoef:
User Story: As a supervisor, I should be able to search forms and students, so that I can easily access the records or forms I need. 

Acceptance Criteria: 
- A supervisor should be able to access the form search and student search that Admins currently have access to.
- If they are a supervisor they should have access limited to the students in their department.
- Admin permissions should remain the same.

Sub-issue of #359

tylerpar99:
Merged.

Issue #359, opened 2022-08-22
Overhaul LSF


AndersonStettner:
### Task List:
- #362 
- #363 
- #365
- #366

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Overhaul Supervisor portal #359](https://trello.com/c/nMLlBh07/509-overhaul-supervisor-portal-359)

PR #383, opened 2022-09-19
Allow Supervisors Limited Student and Form Search Capabilities

bledsoef:
Issue #366 sub-issue of #359 Epic.

## Changes
- Allowed supervisors access to the form search and student search pages and moved those out of the administrator's dropdown menu.
- Created a new logic file which houses student search logic and created the function, limitSearch, which takes in a list of users that were going to be displayed for the student search, and filtered out all students that the current user does not have access to.
- Tested limitSearch in new test file test_search.
- Adjusted queries and conditionals in the main/formSearch.py controller which made it so a user can only search for forms in their department and with other supervisors or students in those departments.

## How to Test
- As an admin, click on the form search and student search pages and confirm that the administrators can see all students, departments and supervisors.
- Then switch to a supervisor with one to a few departments (bolsters is a good one as she just is attached to math) and confirm that their search results are limited only to the scope of their department.
- Run tests.


app/controllers/main_routes/formSearch.py
```diff
@@ -1,4 +1,4 @@
-from app.controllers.admin_routes import admin
+from app.controllers.main_routes import *
```
BrianRamsay: let's import only the symbols we need

app/controllers/main_routes/formSearch.py
```diff
@@ -22,30 +22,54 @@
...
-    students = Student.select().order_by(Student.FIRST_NAME.asc())
+
+    if currentUser.isLaborAdmin or currentUser.isFinancialAidAdmin or currentUser.isSaasAdmin:
+        departments = Department.select().order_by(Department.DEPT_NAME.asc())
+        supervisors = Supervisor.select().order_by(Supervisor.FIRST_NAME.asc())
+        students = Student.select().order_by(Student.FIRST_NAME.asc())
+
+    else:
+
+        departments = (Department.select(Department.DEPT_NAME)
```
BrianRamsay: we should probably select the whole Department record so that our results match up with the query on line 38

app/controllers/main_routes/formSearch.py
```diff
@@ -22,30 +22,54 @@
...
+                            .join_from(Supervisor, LaborStatusForm)
+                            .join_from(LaborStatusForm, Department)
+                            .where(Department.DEPT_NAME.in_(departments))
+                            .distinct())
+
+        students = (Student.select()
+                    .join_from(Student, LaborStatusForm)
+                    .join_from(LaborStatusForm, Department)
+                    .where(Department.DEPT_NAME.in_(departments))
+                    .distinct())
```
BrianRamsay, line 57: These queries should be ordered the same way that the existing ones are (IMO)

app/controllers/main_routes/search.py
```diff
@@ -1,42 +1,36 @@
-from app.controllers.admin_routes import admin
+from app.controllers.main_routes import *
```
BrianRamsay: same as above. import the symbols you need

app/controllers/main_routes/search.py
```diff
@@ -59,4 +53,8 @@ def search(query=None):
     students = list({v['bnumber']:v for v in (current_students + our_students)}.values())
     students = sorted(students, key=lambda f:f['firstName'] + f['lastName'])
 
-    return jsonify(students)
+    if currentUser.isLaborAdmin or currentUser.isFinancialAidAdmin or currentUser.isSaasAdmin:
+        print("ADMINS")
```
BrianRamsay: remove debug output

app/logic/search.py
```diff
@@ -0,0 +1,40 @@
...
+    """ Given a list of student dictionaries and the currentUser the function will only return a
+    list of students who have held a labor position within the currentUsers department."""
+    newstudents = []
+    # Query returns a list of all departments the currentUser has either been a supervisor or created an LSF for.
+    departments = list(Department.select(Department.departmentID)
+                    .join_from(Department, LaborStatusForm)
+                    .join_from(LaborStatusForm, FormHistory)
+                    .where((LaborStatusForm.supervisor == currentUser.supervisor.ID) | (FormHistory.createdBy == currentUser))
+                    .distinct()
+                    )
```
BrianRamsay: This query is essentially the same as the new one in form search. Let's bring it out into its own function so we can a) test it and b) re-use it.

app/logic/search.py
```diff
@@ -0,0 +1,40 @@
...
+    # Query returns a list of student objects for every student who has worked in the departments returned by the "departments" query.
+    students_in_department = list(Student.select(Student.ID)
+                .join_from(Student, LaborStatusForm)
+                .join_from(LaborStatusForm, Department)
+                .where(Department.departmentID.in_(departments))
+                .distinct())
+    studentBnumbers = [student.ID for student in students_in_department]
+    for student in students:
+        if student['bnumber'] in studentBnumbers:
+            newstudents.append(student)
```
BrianRamsay: let's make this iteration either a comprehension or use a function like `filter()`. More pythonic, and matches line 23

app/templates/main/formSearch.html
```diff
@@ -48,17 +49,33 @@ <h1>{{title}}</h1>
...
+                  <!-- We want to either select the first option, or the option that matches the department -->
+                  {% if currentUser.isLaborAdmin or currentUser.isFinancialAidAdmin or currentUser.isSaasAdmin %}
+                      {% if loop.index == 1 or department.DEPT_NAME == department %}
+                        <option value="{{department.DEPT_NAME}}"
+                                selected>{{department.DEPT_NAME}}
+                        </option>
+                      {% else %}
+                        <option value="{{department.DEPT_NAME}}">
+                                      {{department.DEPT_NAME}}
+                        </option>
```
BrianRamsay: instead of repeating the whole `<option>` element, just have the `if` around "selected"

BrianRamsay (changes requested):

AndersonStettner (changes requested):
The admin view dropdowns looked fine but if I hit search when all of the drop downs were in their initial state I was thrown this ajax error: 
image 

Then, when I changed to a supervisor (I used your recommendation of bolsters) there was nothing in any of the dropdowns and when I clicked search I got thrown the same ajax error.

AndersonStettner (changes requested):
Department dropdown is not adding in the data correctly.

AndersonStettner (approved):

app/controllers/main_routes/formSearch.py
```diff
@@ -22,30 +23,53 @@
...
         return render_template('errors/403.html'), 403
 
     terms = LaborStatusForm.select(LaborStatusForm.termCode).distinct().order_by(LaborStatusForm.termCode.desc())
-    departments = Department.select().order_by(Department.DEPT_NAME.asc())
-    supervisors = Supervisor.select().order_by(Supervisor.FIRST_NAME.asc())
-    students = Student.select().order_by(Student.FIRST_NAME.asc())
 
+    if currentUser.isLaborAdmin or currentUser.isFinancialAidAdmin or currentUser.isSaasAdmin:
+        departments = Department.select().order_by(Department.DEPT_NAME.asc())
+        departments = [department for department in departments]
```
BrianRamsay: If you're just turning it into a list use `list()`

app/controllers/main_routes/formSearch.py
```diff
@@ -22,30 +23,53 @@
...
+        supervisors = Supervisor.select().order_by(Supervisor.FIRST_NAME.asc())
+        students = Student.select().order_by(Student.FIRST_NAME.asc())
+
+    else:
+
+        departments = list(getDepartmentsForSupervisor(currentUser))
+
+        # convert department objects to strings
+        departments = [department for department in departments]
```
BrianRamsay: This line isn't doing anything (as it is already a list) and the comment is wrong.

app/controllers/main_routes/search.py
```diff
@@ -1,42 +1,36 @@
...
-    if not currentUser or not currentUser.isLaborAdmin:
+    if not currentUser or not currentUser.supervisor:
         return render_template('errors/403.html'), 403
 
     current_students = []
     our_students = []
     query = query.strip()
 
     # bnumber search
+    # '[Bb]\d+' == is the search a bnumber
```
BrianRamsay: This comment isn't very readable. I would just use english to explain it rather than repeat the confusing code

app/static/js/formSearch.js
```diff
@@ -25,6 +25,7 @@ function runformSearchQuery() {
 
   var termCode = $("#termSelect").val();
   var departmentID = $("#departmentSelect").val();
+  console.log(departmentID)
```
BrianRamsay: Debug output

app/logic/search.py
```diff
@@ -0,0 +1,55 @@
+from app.models.student import Student
+from app.models.supervisor import Supervisor
+from app.models.laborStatusForm import LaborStatusForm
+from app.models.department import Department
+from app.models.formHistory import FormHistory
+
+def limitSearch(students, currentUser):
```
BrianRamsay: This should have a clearer name, like `limitSearchByUserDepartment` or something

BrianRamsay (changes requested):

BrianRamsay (approved):

BrianRamsay (approved):

PR #384, opened 2022-09-22
Fixed dates showing when user leaves term management page

### LINKED ISSUE #379: Term start and end date are blank in Term Management
The start and end dates can be set in Term Management but show up as blank when the page is revisited.

martinj2-dot:
Dates will now appear in the start and end dates fields when a user returns to the term management page in admin settings.


sreynit02 (approved):
Looks good!

Issue #362, opened 2022-09-02
Rework Supervisor Portal


bledsoef:
User Story: As a Supervisor, I want to view and manage department labor students efficiently, so that I can maintain labor student records effectively. 

Acceptance Criteria: 

 - Should be able to easily differentiate labor students by the term they were hired in.
 - Should only have one page to manage all Labor students in my department. 

Note:
 - Admins should still be able to access all departments from the drop-down.

Sub-issue of #359

bledsoef:
We are converting Form Search to the new supervisor portal so we are getting rid of the existing supervisor portal. 
- For each result on Form Search, have a direct link to Labor History
- Make the Form Search an accordion so after searching it is collapsible and you can go right to the table.
- Rename to supervisor portal

tylerpar99:
Merged.

PR #385, opened 2022-09-30
Supervisor UI enhancements 364

bledsoef:


bledsoef:
No longer necessary due to the changes we made to #362

Issue #392, opened 2022-11-21
Turn the Notes Character fields into Text fields and remove character limit on the text fields


AndersonStettner:

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Turn the Notes Character fields into Text fields and remove character limit on the text fields #392](https://trello.com/c/n2G7sCSa/576-turn-the-notes-character-fields-into-text-fields-and-remove-character-limit-on-the-text-fields-392)

PR #399, opened 2023-01-18
changed charfield to textfield and removed character limits

gahimbaref:
#392 
- Changed charfield to textfield for notes model
- Removed placeholder and character limit

Test:
- Go to pending forms under administration, click "actions" and view notes
- You should be able to add notes without limits
- There should be no placeholder

### DIFF
--- a/app/models/notes.py
+++ b/app/models/notes.py
@@ -6,6 +6,6 @@ class Notes(baseModel):
     noteHistoryID       = PrimaryKeyField()
     formID              = ForeignKeyField(LaborStatusForm, on_delete="cascade")               # foreign key to lsf
     createdBy           = ForeignKeyField(User, on_delete="cascade")
-    notesContents       = CharField()
+    notesContents       = TextField()
     noteType            = CharField()
     date                = DateField()
--- a/app/templates/admin/financialAidOverload.html
+++ b/app/templates/admin/financialAidOverload.html
@@ -84,7 +84,7 @@
        </div>
        <div class="form-group">
          <label>{{userDept}} Note: </label>
-         <textarea type="text" class="form-control" id="{{lsfForm.laborStatusFormID}}" maxlength="250" placeholder="250 Characters Maximum"></textarea>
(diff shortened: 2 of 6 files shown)


Karina-Agliullova:
When I save a note that is too long, and then I click on 'actions' and view the notes, it goes over the borders
Screenshot (1291)

Karina-Agliullova:
Also, I think having a slightly bigger area for text in notes would be helpful because it seems to be more user-friendly

sreynit02:
> When I save a note that is too long, and then I click on 'actions' and view the notes, it goes over the borders Screenshot (1291)

This was an issue from before. It happens because the word is too long. Normally, we would not have a word that is that long make this error. We consulted Anderson and he said it is ok.

sreynit02:
> Also, I think having a slightly bigger area for text in notes would be helpful because it seems to be more user-friendly

I think the size of the textbox is ok since it is located on a modal which is already small. We would have to make the modal bigger if we want a bigger textbox or take over space that could be the note log. Plus the notes are generally short.

AndersonStettner (changes requested):
nevermind

AndersonStettner:
> > Also, I think having a slightly bigger area for text in notes would be helpful because it seems to be more user-friendly
> 
> I think the size of the textbox is ok since it is located on a modal which is already small. We would have to make the modal bigger if we want a bigger textbox or take over space that could be the note log. Plus the notes are generally short.

If you increase the amount of rows when you create the textarea it will increase the height of the textarea and the modal will adjust itself to the size of the textarea by default.

AndersonStettner:
> When I save a note that is too long, and then I click on 'actions' and view the notes, it goes over the borders Screenshot (1291)

If you give the div around the note these styles, it will keep super-long words from going outside of the box.
```
    text-overflow: ellipsis;
    white-space: nowrap;
    width: 100%;
    overflow: hidden;
```

solijonovam:
looks all good! @Andersonstettner

AndersonStettner (approved):

PR #400, opened 2023-01-19
updated lsf when modifying hours

sreynit02:
- Updated fields in the Labor Status Form to reflect adjustments done to approved forms. 

Test: 
- Create a Labor Status Form for a 10 hr position for a student and approve it. Change the 'supervisor' field. Also change the hours from 10 to 15. 
- Check the Labor Status Form. The supervisor entry should change to new supervisor. The weeklyHours field should be 20 instead of 10. There should also be an overload form submitted for this position as well.

### DIFF
--- a/app/controllers/main_routes/alterLSF.py
+++ b/app/controllers/main_routes/alterLSF.py
@@ -246,9 +246,17 @@ def adjustLSF(fieldsChanged, fieldName, lsf, currentUser, host=None):
                                            createdBy    = currentUser,
                                            createdDate  = date.today(),
                                            status       = status.statusName)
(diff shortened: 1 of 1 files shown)


Karina-Agliullova:
Screenshot (1289)
Screenshot (1290)
Whenever I try to change a supervisor, it crashes and gets stuck on that page while processing. If I update, it shows an error. Also, it doesn't let me change the hours unless I have the initially assigned supervisor.

sreynit02:
This is no longer a bug. The Labor Status Form only get updated when the adjustment form has been approved.

PR #401, opened 2023-01-23
Update students tracy 371

solijonovam:
Summary:
Created 2 new functions: `updateStudentRecords` will update students DB records to whatever is in Tracy, `updateSupervisorRecord` does the same function for supervisors. Then we created a script that when called will update students and supervisors in our DB. 

To test:
Run test suite, and in staging run the script and you will see some records update and some don't (old records).

### DIFF
--- a/app/logic/updateStudentData.py
+++ b/app/logic/updateStudentData.py
@@ -1,0 +1,6 @@
+from app.logic.userInsertFunctions import updateStudentDBRecords
+
+def runUpdateFunction():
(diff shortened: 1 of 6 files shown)


Karina-Agliullova (approved):
After many trials and errors, we were able to test it and it worked! Good job!

sreynit02 (approved):
Looks good!

BrianRamsay (approved):

Issue #378, opened 2022-09-16
Update departments with new org/account codes


BrianRamsay:
We need to periodically update departments from Tracy data, as org and account codes can change.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Update departments with new org/account codes #378](https://trello.com/c/mWoTErD1/558-update-departments-with-new-org-account-codes-378)

PR #406, opened 2023-02-21
Update departments account and org from Tracy.

tylerpar99:
This fixed PR #378 

Summary: This PR changes the file `updateStudentRecords.py` to `updateDBRecords.py`. We then add to that file a count of failed and updated department codes. In `userInsertFunctions.py` we add a new function that will update all department account and org codes based on what Tracy has for that department name. We also added a new function to Tracy to get department data from a department name and return the account and org code. We also added to the test suite to include the new department updating function and updated a test that resulted from adding a new department to UTE.

To Test: Run the test suite. Reset from-backup and export flask environment to staging. Not in the terminal type "python updateDBRecords.py" and you will see that some fail and some pass. Fails are from departments that no longer exist. Success proves that the new function updates departments.


app/logic/userInsertFunctions.py
```diff
@@ -52,28 +53,45 @@ def createUser(username, student=None, supervisor=None):
 
 def updateDBRecords():
     """
-    This function will update all student and supervisor records according to
+    This function will update all student, supervisor and department records according to
     Tracy data.
     """
     studentsInDB = Student.select()
     supervisorsInDB = Supervisor.select()
+    departmentsFromDB = Department.select()
```
AndersonStettner: Not a blocker and really nitpicky but I think it would be nice to keep naming the same on all of these so it would be departmentsInDB

sreynit02: agreed!

sreynit02 (changes requested):
Just one easy comment to resolve. After that you're all set!

app/logic/tracy.py
```diff
@@ -91,6 +91,13 @@ def getDepartments(self):
         return STUPOSN.query.with_entities(STUPOSN.ORG, STUPOSN.DEPT_NAME, STUPOSN.ACCOUNT) \
                             .distinct().order_by(STUPOSN.DEPT_NAME).all()
 
+    def getDepartmentsFromName(self, deptName):
+        """
```
stevensonmichel, line 95: You might consider correcting the name of the function. getDepartmentsFromName implies that you are getting multiple departments, which is not true.

BrianRamsay:
Implemented elsewhere in the last 2 years

PR #411, opened 2023-04-20
Preferred name

BrianRamsay:


AndersonStettner (changes requested):
The database reset fails. The last line of the error is `AttributeError: 'property' object has no attribute 'get_sort_key'`

image

Issue #381, opened 2022-09-16
Form creation should have fewer confirmations


BrianRamsay:
The Add Student / Review process when adding forms should be updated to include fewer confirmations. Instead, users should add students and have editing and withdrawal options rather than have to click several times to create a form.

Many users are not aware that Add Student does not actually add a form for the student.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Form creation should have fewer confirmations #381](https://trello.com/c/TME7myXX/560-form-creation-should-have-fewer-confirmations-381)

PR #413, opened 2023-04-26
Fewer form confirmation 381

tylerpar99:
This pr fixes issue #381 

Summary: 
  The previous LSF creation page had a series of confirmations that seemed redundant to users. 

- This PR has condensed the confirmations to two clicks, an "Add Student" button that will trigger a confirmation modal. 
- We have also changed the old table to a data table that will repopulate with all of the current users created and assigned forms. 
- Lastly, we changed the way cooked were being used on this page so that they will keep the fields populated after a form is created for easier batch adding, we also made sure the cookies from this page and the supervisorPortal work together. 

To Test: 
- Set the flask environment to staging, upon page load, you should see forms pre-loaded into the table below. You can then fill out a form for a student and see that after fully adding the form most of the fields will be populated for another add if you choose to do so. 
- If you go to the supervisor portal page and complete a search you should see that the cookies from both pages don't interfere with one another. 
- Run tests suite and make sure everything passes.


tylerpar99:
The only thing left to do is figure out how to get the position selector to fill from the cookie's data. We tried what we believe to be every combination stack overflow has to offer and have yet to get it successfully fill. My suspicion is that the selector options aren't filling in time and we are essentially attempting to select from an empty list. Brian is looking into it.

bledsoef:
[Form creation should have fewer confirmations #381](https://trello.com/c/TME7myXX/560-form-creation-should-have-fewer-confirmations-381)

PR #426, opened 2023-09-06
423 inactive supervisors

### LINKED ISSUE #423: Separate Inactive supervisors from Active in form search
In the supervisor dropdown in the form search, inactive supervisors should be visually distinct and sorted to the bottom.

hoerstl:
**Issue** - Some supervisors are considered inactive by labor status standards. A supervisor is inactive if they are not present within the Tracy database but you are still in our DB from the past. Since these inactive supervisors are less likely to be selected, we want them to be sorted to the bottom of the form search and visually distinct for ease of use.

**Solution** - We added an isActive column to the supervisor table so that we could check whether or not they are active in the Tracy system or not. This column is populated False by default and is only populated True by a particular Python script that we already run in the middle of the night (a file we have renamed in this pr "updateStudentRecords.py" -> "updateDBRecords.py" since it already updated supervisors too). This file checks the Tracy DB for each user and if they are present, marks them as active in our system. Additionally, there's a SQL script we need to run that fixes some supervisor names that have their whole name set as their LAST_NAME.

Later, when we go to retrieve the supervisors, we order them by their activity, then their preferred name, and then their legal name so that the inactive persons appear at the bottom of the list in alphabetical order.

**To Test** - First we need to get your DB in order. Follow the commands below:
1. source setup.sh
2. database/reset_database.sh from-backup
3. export FLASK_ENV=staging
4. python scripts/updateDBRecords.py 
(This is the file that checks Tracy in the middle of the night to mark people as active or inactive)
5. mysql -u lsf_user -p lsf < scripts/fixIncorrectNames.sql
(The password is "password". This is the script that fixes the incorrect LAST_NAME s of the 6 supervisors)
6. flask run

Now your DB is all set up with the new column we added in prod-backup.sql and updated by running the updateDBRecords.py file. 
To test, open the application and go to the form search.
image

Ensure the results are ordered with inactive at the bottom and mostly in alphabetical order and that the visual indication of inactivity is visually pleasing.
(There are two people who have no first name in prod-backup and their entire name is contained within the last name field. Therefore when we sort them, since the shortest name is alphabetically first, they come at the front of the inactive folk.)

image

Resolves #423


solijonovam:
the order of the names for inactive supervisors are being sorted 2 times.

solijonovam (changes requested):
ACTIVE:
there is a short list of some sort of supervisors, and it stops at 'T,' and then re-sorts again a bit more properly.
INACTIVE:
the order of the names for inactive supervisors are being sorted 2 times.

AndersonStettner (commented):
Make sure you are branching off of development when you start working on an issue. This branch has changes from branch `418-Easy-Email`.

hoerstl:
> ACTIVE: there is a short list of some sort of supervisors, and it stops at 'T,' and then re-sorts again a bit more properly. INACTIVE: the order of the names for inactive supervisors are being sorted 2 times.

I see! I found the issue. We thought that sorting by preferred name and then legal name was enough but whenever someone didn't have one of those, they were sorted first alphabetically instead of not being sorted specially at all which split the list into two.

andersoncedu:
This looks good to me.

app/controllers/main_routes/main_routes.py
```diff
@@ -47,30 +47,35 @@ def supervisorPortal():
     
     terms = LaborStatusForm.select(LaborStatusForm.termCode).distinct().order_by(LaborStatusForm.termCode.desc())
     allSupervisors = Supervisor.select()
+    supervisor_first_name = fn.COALESCE(Supervisor.preferred_name, Supervisor.legal_name)
+    student_first_name = fn.COALESCE(Student.preferred_name, Student.legal_name)
```
AndersonStettner: We should be using camelCase for consistency.

AndersonStettner (changes requested):

app/logic/userInsertFunctions.py
```diff
@@ -127,6 +128,7 @@ def updateSupervisorRecord(supervisor):
     supervisor.CPO = tracyUser.CPO
     supervisor.ORG = tracyUser.ORG
     supervisor.DEPT_NAME = tracyUser.DEPT_NAME
+    supervisor.isActive = bool(tracyUser)
```
BrianRamsay: This should be set to `True`, as it only gets to this point when tracyUser is valid. Using `bool(tracyUser)` implies that it could potentially be false, which is incorrect.

app/controllers/main_routes/main_routes.py
```diff
@@ -47,30 +47,35 @@ def supervisorPortal():
...
+    studentFirstName = fn.COALESCE(Student.preferred_name, Student.legal_name)
     if currentUser.isLaborAdmin or currentUser.isFinancialAidAdmin or currentUser.isSaasAdmin:
         departments = Department.select().order_by(Department.DEPT_NAME.asc())
         departments = [department for department in departments]
-        supervisors = Supervisor.select().order_by(Supervisor.preferred_name, Supervisor.legal_name)
-        students = Student.select().order_by(Student.preferred_name, Student.legal_name)
+        supervisors = (Supervisor.select(Supervisor, supervisorFirstName.alias('first_name'))
+                                 .order_by(Supervisor.isActive.desc(), supervisorFirstName))
+        students = (Student.select(Student, studentFirstName.alias('first_name'))
+                           .order_by(studentFirstName))
```
BrianRamsay: We don't ever use these coalesced names, do we? In your html you use the pre-existing FIRST_NAME property on the student and supervisor. We should just be able to drop the arguments to `select()`

BrianRamsay (changes requested):

app/logic/userInsertFunctions.py
stevensonmichel: Fixed

app/controllers/main_routes/main_routes.py
stevensonmichel: Fixed

AndersonStettner (changes requested):
Nevermind

app/controllers/main_routes/main_routes.py
```diff
@@ -47,30 +47,36 @@ def supervisorPortal():
...
     else:
 
         departments = getDepartmentsForSupervisor(currentUser)
 
         # convert department objects to strings
-        departments = [department for department in departments]
         deptNames = [department.DEPT_NAME for department in departments]
 
-        supervisors = (Supervisor.select()
+        supervisors = (Supervisor.select(Supervisor, supervisorFirstName.alias('first_name'))
```
AndersonStettner: There is no need for this alias and it is never used.

AndersonStettner (changes requested):

app/controllers/main_routes/main_routes.py
stevensonmichel: Fixed

AndersonStettner (changes requested):
All "Unknown User" users should be moved to the end of the list in the supervisor dropdown in the Form Search section

AndersonStettner (approved):

PR #428, opened 2023-09-12
425 inactive departments

### LINKED ISSUE #425: Split inactive and active departments in Department Management
Departments in the Department Management page should be split into two tabs, one for Inactive and one for Active. Inactive means the department is not in the Tracy data for positions (STUPOSN).

hoerstl:
Issue - There are some departments that we would consider inactive within our system. The definition for being inactive is that these departments are in our local database but they have been removed from the Tracy DB. As an enhancement, we wanted to split these inactive departments into two different tabs on the administrative "Manage Departments" page for ease of use.

Solution - We added a column, isActive, into the database for the department table. There is a script run once during the night on our application which is named updateStudentRecords.py which now checks for the most current info on all personnel and departments from Tracy. By checking once daily we ensure that the departments are correctly marked as active or inactive. In the route for the "Manage Departments" page, we simply query for the active departments and inactive departments separately before loading them into two different tabs on the page. 

To Test - Follow these steps:
1. Run `source setup.sh`
2. Reset your DB `database/reset_database.sh from-backup`
3. Change your environment to staging `export FLASK_ENV=staging`. This step ensures that the following script checks the real Tracy DB instead of the development one.
4. Run the updateStudentRecords.py script, `python scripts/updateDBRecords.py`. Be sure to wait until this completes.
5. Run the application `flask run`.
6. Navigate to the Manage Departments page by clicking Administration on the left-hand side of the window.

How you know it's good:
Ensure that there are departments in both the active and inactive departments tabs. Additionally, you should also ensure that the tabs function as you believe tabs should (shifting focus on click, that kind of thing). There should be some departments in both. For us, there were 143 active and 44 inactive departments. 

Resolves #425


app/static/js/manageDepartments.js
```diff
@@ -2,11 +2,30 @@
 $("#admin").collapse("show");
 
 $(document).ready( function(){
-    x = $('#departmentsTable');
+    x = $('#activeDepartmentsTable');
```
qasema: It would be better to use descriptive variables instead of x and y

qasema (changes requested):

app/static/js/manageDepartments.js
hoerstl: I agree, changes have been made

qasema (approved):
looks good

solijonovam:
i'm not sure if this is an issue, but when I run the Python script it's saying that there are 44 departments not found, but when I go on the page I am finding 47 inactive departments.

AndersonStettner (changes requested):
The first time the `Manage Departments` page is loaded both the active and inactive tables are on the same page. 

The name of the script, `updateStudentRecords.py`, is quite misleading. In the context of this PR everything that is being done is in relation to departments, then if you look at the script itself it is dealing with students, supervisors, and departments.

hoerstl:
> i'm not sure if this is an issue, but when I run the Python script it's saying that there are 44 departments not found, but when I go on the page I am finding 47 inactive departments.

Odd. When I run the script, I get 44 departments not found and then the inactive tab shows 44 department entries. I'll have to talk to you about that

AndersonStettner (dismissed):
Figure out why there are inconsistencies in number of departments not found by the script and number of inactive departments in the application.

hoerstl:
> Figure out why there are inconsistencies in number of departments not found by the script and number of inactive departments in the application.

Figured it out. We were fetching new departments from Tracy whenever we pinged the manage departments page instead of when we ran the update script. This was causing new and existing departments in Tracy to be marked as inactive when they should be. Good catch you guys!

andersoncedu:
This looks good to me.

AndersonStettner (changes requested):
Look into why we have duplicates of active departments. 
image

Also, if the department is "Inactive" it should no longer have an `In Compliance` or `Not in Compliance` button.

AndersonStettner (changes requested):
There are inactive departments showing in the department drop-down on the form search. According to Brian, we should only show active departments here.

hoerstl:
> There are inactive departments showing in the department drop-down on the form search. According to Brian, we should only show active departments here.

From a conversation with Brian -- Correction, we are showing all departments (active or no) on the form search and we are to grey them out as they are in the inactive supervisors branch. It's the form creation page where we should only show active departments.

AndersonStettner (approved):

Issue #417, opened 2023-08-21
Adjustment forms are not going through properly when the adjustment triggers an overload.


AndersonStettner:
First reproduce the issue, find the problem, then debug.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Adjustment forms are not going through properly when the adjustment triggers an overload. #417](https://trello.com/c/rtq3wfKe/685-adjustment-forms-are-not-going-through-properly-when-the-adjustment-triggers-an-overload-417)

PR #429, opened 2023-09-18
417 adjustment error

hoerstl:
Issue: There was a bug that occurred whenever anyone made an adjustment to a labor form that would mean they were overloading and would stop the adjustment from going through. FYI, overloading means to work strictly greater than 15 hours per week. For example, if someone was working a primary for 10 hours and had a pending secondary form for 5 hours and we adjusted it to be a secondary for 10 hours, our server logic would crash and the form wouldn't go through.

Solution: We found out that this issue was simply due to a misunderstanding of some variable names within the createOverloadForm function. The parameter "newWeeklyHours" was being passed in as the new number of hours to change a job to, but was being interpreted inside the function as the total number of hours that the user was now working/the number of hours they are working in addition to their current schedule. Needless to say this logic was wrought with inconsistency so we rewrote the function to alter the form, gathering the correct total number of hours before and after the change, using that to correctly create and delete overload forms as necessary.

To test: 
Assuming you're in the lsf directory
1. source setup.sh
2. database/reset_database.sh from-backup
3. export FLASK_ENV=staging
4. flask run
Now open the application and click Administration->Pending Forms
image

There should be a list of pending forms. Keep searching for students until you find someone who is not currently overloading for a single academic year. If they have not overloaded, then click Actions->Modify on it and change the hours to overloading. If the request goes through and you are redirected back to the 'pending forms' page, then the fix was successful. Now try to undo the process. Go to the same person and alter it back to ensure that there are no crashes when removing the overload form.

Resolves issue #417

### DIFF
--- a/app/controllers/main_routes/alterLSF.py
+++ b/app/controllers/main_routes/alterLSF.py
@@ -142,10 +142,8 @@ def submitAlteredLSF(laborStatusKey):
         currentDate = datetime.now().strftime("%Y-%m-%d")
         fieldsChanged = eval(request.data.decode("utf-8")) # This fixes byte indices must be intergers or slices error
         fieldsChanged = dict(fieldsChanged)
-        print(fieldsChanged)
         student = LaborStatusForm.get(LaborStatusForm.laborStatusFormID == laborStatusKey)
         formStatus = (FormHistory.get(FormHistory.formID == laborStatusKey).status_id)
-        print(formStatus)
         formHistoryIDs = []
         lsf = LaborStatusForm.get(LaborStatusForm.laborStatusFormID == laborStatusKey)
         for fieldName in fieldsChanged:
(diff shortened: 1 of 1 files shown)


gahimbaref:
looks good!

doucoureb:
Looks good!

AndersonStettner (commented):
- I found a student, `Bashiri Sawyer`, who only had a primary position in for `AY 2023-2024`. 
- I created a 5 hour secondary position for the student and, as expected, no overload form was required. 
- I then modified the new secondary to be a 10 hour position and an overload was correctly required. 
- Finally, I tried to adjust the hours from 10 back down to 5. After doing this I received messaging indicating that my change was made successfully but the hours were still at 10 and not 5. 

After looking into it a bit, it seems that the second update is not being saved at all since even the note that is left when modifying 10 hours back down to 5 hours is not saved. Investigate to see if this is caused by one of your changes.

AndersonStettner (commented):
If this is ready for re review please remove the "TODO" and print statements.

hoerstl:
> If this is ready for re review please remove the "TODO" and print statements.

You know, that's a darn good point.

AndersonStettner (approved):

PR #430, opened 2023-09-19
424 supervisor department

### LINKED ISSUE #424: Admin management of department members
The admins should have a way to see the supervisors associated with a department and remove ones that have been added.

From the Manage Departments page, there should be a link or button that opens a modal. The modal should have a dropdown that can add a supervisor to a department (like the one in the supervisor portal), but also a list of the current supervisors of the department. The ones who have been added to the department (ie, not there from forms), should have a Remove button.

andersoncedu:
Issue: There was a request to add a modal on the manage department section of the website which would be able to add and remove supervisors assigned to a particular department.

Fix: Created a modal that pulls all the entries associated with a department from the supervisorDepartment table and displays them for a particular department along with a remove button that will remove that particular entry from the table. Also added a dropdown list to show all supervisors and allow you to select a supervisor to add. 

Test: To navigate to this modal, from the sidebar click the administration dropdown and click manage department. From there, you should be able to select and of the department name and attach and remove supervisors.

Fixes #424

### DIFF
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -32,7 +32,7 @@ def manage_departments():
                 return render_template('errors/403.html'), 403
 
         departmentTracy = Tracy().getDepartments()
-        allSupervisors= Supervisor.select()
+        allSupervisors= Supervisor.select().order_by(Supervisor.LAST_NAME)
         for dept in departmentTracy:
             d, created = Department.get_or_create(DEPT_NAME = dept.DEPT_NAME, ACCOUNT=dept.ACCOUNT, ORG = dept.ORG)
             d.save()
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -47,7 +47,6 @@ def manage_departments():
 
 @admin.route("/admin/manageDepartments/<departmentID>", methods=['GET'])
 def getSupervisorsInDepartment(departmentID):
-    try:
         currentUser = require_login()
         if not currentUser:                    # Not logged in
             return render_template('errors/403.html')
@@ -58,11 +57,9 @@ def getSupervisorsInDepartment(departmentID):
                 return render_template('errors/403.html'), 403
         
         supervisors = getSupervisorsForDepartment(departmentID)
-        departmentID=Department.get_by_id(departmentID)
-        supervisors= [model_to_dict(supervisor) for supervisor in supervisors]
-        return jsonify([model_to_dict(departmentID), supervisors])
-    except Exception as e:
-        return render_template('errors/500.html'), 500
+        selectedDepartment = Department.get_by_id(departmentID)
+        supervisors = [model_to_dict(supervisor) for supervisor in supervisors]
+        return jsonify([model_to_dict(selectedDepartment), supervisors])
     
 @admin.route('/admin/manageDepartments/removeSupervisorFromDepartment', methods=['POST'])
 def removeSupervisorFromDepartment():
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -47,7 +47,7 @@ def manage_departments():
         return render_template('errors/500.html'), 500
 
 @admin.route("/admin/manageDepartments/<departmentID>", methods=['GET'])
-def getSupervisorsinDepartment(departmentID):
+def getSupervisorsInDepartment(departmentID):
     try:
         currentUser = require_login()
         if not currentUser:                    # Not logged in
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -1,6 +1,7 @@
 from app.controllers.admin_routes import *
 from app.models.user import *
 from app.login_manager import require_login
+from app.logic.search import getSupervisorsForDepartment
 from app.controllers.admin_routes import admin
 from app.controllers.errors_routes.handlers import *
 #from app.models.manageDepartments import *
@@ -9,6 +10,7 @@ from flask_bootstrap import bootstrap_find_resource
 from app.models.department import *
 from flask import request, redirect
 from flask import jsonify
+from playhouse.shortcuts import model_to_dict
 from app.logic.tracy import Tracy
 
 @admin.route('/admin/manageDepartments', methods=['GET'])
@@ -41,6 +43,27 @@ def manage_departments():
         print("Error Loading all Departments", e)
         return render_template('errors/500.html'), 500
 
+@admin.route("/admin/manageDepartments/<currentDepartment>", methods=['GET'])
+def getSupervisorsinDepartment(currentDepartment):
+    print("/"*100)
+    try:
+        currentUser = require_login()
+        if not currentUser:                    # Not logged in
+            return render_template('errors/403.html')
+        if not currentUser.isLaborAdmin:       # Not an admin
+            if currentUser.student: # logged in as a student
+                return redirect('/laborHistory/' + currentUser.student.ID)
+            elif currentUser.supervisor:
+                return render_template('errors/403.html'), 403
+        
+        supervisors = getSupervisorsForDepartment(currentDepartment)
+        currentDepartment=Department.get_by_id(currentDepartment)
+        print("/"*1000)
+        supervisors= [model_to_dict(supervisor) for supervisor in supervisors]
+        print(supervisors)
+        return jsonify([model_to_dict(currentDepartment), supervisors])
+    except Exception as e:
+        return render_template('errors/500.html'), 500
 
 @admin.route('/admin/complianceStatus', methods=['POST'])
 def complianceStatusCheck():
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -46,8 +46,8 @@ def manage_departments():
         print("Error Loading all Departments", e)
         return render_template('errors/500.html'), 500
 
-@admin.route("/admin/manageDepartments/<currentDepartment>", methods=['GET'])
-def getSupervisorsinDepartment(currentDepartment):
+@admin.route("/admin/manageDepartments/<departmentID>", methods=['GET'])
+def getSupervisorsinDepartment(departmentID):
     try:
         currentUser = require_login()
         if not currentUser:                    # Not logged in
@@ -58,17 +58,15 @@ def getSupervisorsinDepartment(currentDepartment):
             elif currentUser.supervisor:
                 return render_template('errors/403.html'), 403
         
-        supervisors = getSupervisorsForDepartment(currentDepartment)
-        currentDepartment=Department.get_by_id(currentDepartment)
+        supervisors = getSupervisorsForDepartment(departmentID)
+        departmentID=Department.get_by_id(departmentID)
         supervisors= [model_to_dict(supervisor) for supervisor in supervisors]
-        return jsonify([model_to_dict(currentDepartment), supervisors])
+        return jsonify([model_to_dict(departmentID), supervisors])
     except Exception as e:
         return render_template('errors/500.html'), 500
     
 @admin.route('/admin/manageDepartments/removeSupervisorFromDepartment', methods=['POST'])
 def removeSupervisorFromDepartment():
-    supervisorDeptRecord=request.form
-    supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = supervisorDeptRecord['supervisor'], department = supervisorDeptRecord['department'])
     try:
         currentUser = require_login()
         if not currentUser:                    # Not logged in
@@ -78,16 +76,18 @@ def removeSupervisorFromDepartment():
                 return redirect('/laborHistory/' + currentUser.student.ID)
             elif currentUser.supervisor:
                 return render_template('errors/403.html'), 403
-            
+        
+        supervisorDeptRecord = request.form
+        supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = supervisorDeptRecord['supervisorID'], department = supervisorDeptRecord['departmentID'])
+    
         if supervisorDeptRecord:
             supervisorDeptRecord.delete_instance()
             return "True"
-
         else:
             return "False"
     
     except Exception as e:
-        print(f'Could not add user to department: {e}')
+        print(f'Could not remove user from department: {e}')
         return "", 500
 
 @admin.route('/admin/complianceStatus', methods=['POST'])
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -70,6 +70,15 @@ def removeSupervisorFromDepartment():
     supervisorDeptRecord=request.form
     supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = supervisorDeptRecord['supervisor'], department = supervisorDeptRecord['department'])
     try:
+        currentUser = require_login()
+        if not currentUser:                    # Not logged in
+            return render_template('errors/403.html')
+        if not currentUser.isLaborAdmin:       # Not an admin
+            if currentUser.student: # logged in as a student
+                return redirect('/laborHistory/' + currentUser.student.ID)
+            elif currentUser.supervisor:
+                return render_template('errors/403.html'), 403
+            
         if supervisorDeptRecord:
             supervisorDeptRecord.delete_instance()
             return "True"
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -77,8 +77,8 @@ def removeSupervisorFromDepartment():
             elif currentUser.supervisor:
                 return render_template('errors/403.html'), 403
         
-        supervisorDeptRecord = request.form
-        supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = supervisorDeptRecord['supervisorID'], department = supervisorDeptRecord['departmentID'])
+        formData = request.form
+        supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = formData['supervisorID'], department = formData['departmentID'])
     
         if supervisorDeptRecord:
             supervisorDeptRecord.delete_instance()
--- a/app/controllers/admin_routes/manage_departments.py
+++ b/app/controllers/admin_routes/manage_departments.py
@@ -1,5 +1,6 @@
 from app.controllers.admin_routes import *
 from app.models.user import *
+from app.models.supervisorDepartment import SupervisorDepartment
 from app.login_manager import require_login
 from app.logic.search import getSupervisorsForDepartment
 from app.controllers.admin_routes import admin
@@ -31,13 +32,15 @@ def manage_departments():
                 return render_template('errors/403.html'), 403
 
         departmentTracy = Tracy().getDepartments()
+        allSupervisors= Supervisor.select()
         for dept in departmentTracy:
             d, created = Department.get_or_create(DEPT_NAME = dept.DEPT_NAME, ACCOUNT=dept.ACCOUNT, ORG = dept.ORG)
             d.save()
         department = Department.select()
         return render_template( 'admin/manageDepartments.html',
                                 title = ("Manage Departments"),
-                                department = department
+                                department = department,
+                                allSupervisors = allSupervisors
                                 )
     except Exception as e:
         print("Error Loading all Departments", e)
@@ -45,7 +48,6 @@ def manage_departments():
 
 @admin.route("/admin/manageDepartments/<currentDepartment>", methods=['GET'])
 def getSupervisorsinDepartment(currentDepartment):
-    print("/"*100)
     try:
         currentUser = require_login()
         if not currentUser:                    # Not logged in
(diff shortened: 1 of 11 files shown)


andersoncedu:
[Admin management of department members](https://trello.com/c/XiX4ibl6/694-admin-management-of-department-members)

app/controllers/admin_routes/manage_departments.py
```diff
@@ -29,18 +32,63 @@ def manage_departments():
...
                                 title = ("Manage Departments"),
-                                department = department
+                                department = department,
+                                allSupervisors = allSupervisors
                                 )
     except Exception as e:
         print("Error Loading all Departments", e)
         return render_template('errors/500.html'), 500
 
+@admin.route("/admin/manageDepartments/<currentDepartment>", methods=['GET'])
```
hoerstl: You make the following change:
Rename `currentDepartment` -> `departmentId` or `departmentID`

hoerstl: Also, be sure to replace it everywhere it appears in this function

app/controllers/admin_routes/manage_departments.py
```diff
@@ -29,18 +32,63 @@ def manage_departments():
...
+        supervisors = getSupervisorsForDepartment(currentDepartment)
+        currentDepartment=Department.get_by_id(currentDepartment)
+        supervisors= [model_to_dict(supervisor) for supervisor in supervisors]
+        return jsonify([model_to_dict(currentDepartment), supervisors])
+    except Exception as e:
+        return render_template('errors/500.html'), 500
+    
+@admin.route('/admin/manageDepartments/removeSupervisorFromDepartment', methods=['POST'])
+def removeSupervisorFromDepartment():
+    supervisorDeptRecord=request.form
```
hoerstl: On lines 70 and 71 (before it's reassigned), `supervisorDeptRecord` would be better named `deletionRequest` since the `request.form` is not the department record.

app/controllers/admin_routes/manage_departments.py
```diff
@@ -29,18 +32,63 @@ def manage_departments():
...
+    supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = supervisorDeptRecord['supervisor'], department = supervisorDeptRecord['department'])
```
hoerstl: This comment goes hand in hand with another in the JavaScript. We should rename the keys `'supervisor'` and `'department'` with `'supervisorId'` and `'departmentId'` respectively.

app/controllers/admin_routes/manage_departments.py
```diff
@@ -29,18 +32,63 @@ def manage_departments():
...
+        if not currentUser.isLaborAdmin:       # Not an admin
+            if currentUser.student: # logged in as a student
+                return redirect('/laborHistory/' + currentUser.student.ID)
+            elif currentUser.supervisor:
+                return render_template('errors/403.html'), 403
+            
+        if supervisorDeptRecord:
+            supervisorDeptRecord.delete_instance()
+            return "True"
+
```
hoerstl: I would delete this empty line here. If-else blocks are usually together when they're short

app/controllers/admin_routes/manage_departments.py
```diff
@@ -29,18 +32,63 @@ def manage_departments():
...
+        else:
+            return "False"
+    
+    except Exception as e:
+        print(f'Could not add user to department: {e}')
```
hoerstl: 'Could not **add user to** department: {e}' -> 'Could not **remove user from** department: {e}'

This is the remove supervisor function after all

app/logic/search.py
```diff
@@ -64,3 +64,7 @@ def getDepartmentsForSupervisor(currentUser):
     # removes duplicate departments
     alldepts = list(set(alldepts))
     return alldepts
+
+def getSupervisorsForDepartment(currentDepartment):
```
hoerstl: Rename `currentDepartment` to `departmentId` for specificity. Otherwise, it might seem like a department object.

app/logic/search.py
```diff
@@ -64,3 +64,7 @@ def getDepartmentsForSupervisor(currentUser):
     # removes duplicate departments
     alldepts = list(set(alldepts))
     return alldepts
+
+def getSupervisorsForDepartment(currentDepartment):
+    supervisors = Supervisor.select().join(SupervisorDepartment).where(SupervisorDepartment.department == currentDepartment).order_by(Supervisor.LAST_NAME)
```
hoerstl: I'd rename `supervisors` to `departmentSupervisors` and I'd execute it on this line instead of following the return statement but the ladder is just my preference.

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -0,0 +1,27 @@
+function addSupervisorToDepartment(data) {
```
hoerstl: Instead of passing in data here, it would be better for this function to have two parameters, a supervisor id and a department id and have the function create data from those. Right now, the parameter data tells us nothing about what format we're supposed to prepare data in or what information it might need. 

After, you'll also need to change your calls to the function to match.

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -0,0 +1,27 @@
...
+        }
+      },
+      error: function() {
+        msgFlash("Failed to add supervisor, please try again.", "fail")
+        clearDropdown()
+      },
+    })
+}
+
+function clearDropdown(){
```
hoerstl: You should rename this function to be plural since the function actually clears the selection for each of the `select`s that have the class `selectpicker`. Be sure to change the spelling in your usage of the function above.

app/controllers/admin_routes/manage_departments.py
AndersonStettner: We should probably move the request.form and supervisor get to after we verify the user below.

AndersonStettner (changes requested):

app/static/js/manageDepartments.js
```diff
@@ -7,8 +7,88 @@ $(document).ready( function(){
...
+      getSupervisorsInDepartment(this.id)
+    })
+}
+
+
+$("#supervisorModalSelect").on('change', async function() {
+  let supervisor = $('#supervisorModalSelect :selected').val()
+  let department = $('#departmentModalSelect').data('department-id')
+  let data = {"supervisor": supervisor, "department": department}
+  await addSupervisorToDepartment(data)
```
hoerstl: This await does not do what you think it does. Recall that the ajax method completes instantaneously after sending out the information to the server. What you want to do is to wait until the data has reached the server successfully but what you're actually doing is only waiting for the ajax to send the information, not to wait until it's been received. This results in an incomplete fetch of the supervisors that should be in the department on the next line since it might exclude the data that we just tried to update. 

To fix this, consider adding a callback parameter to the `addSupervisorToDepartment` function and passing in a function containing everything you'd like to do after the data is received by the server. You can call the callback from the success key-value pair when you make the ajax request.

app/static/js/manageDepartments.js
```diff
@@ -7,8 +7,88 @@ $(document).ready( function(){
...
+      $('#manageDepartmentSupervisorModal').modal('show')
+      $('.removeSupervisorFromDepartment').on('click', removeSupervisorFromDepartment)
+    }
+    })
+  }
+
+function removeSupervisorFromDepartment () {
+  let department = $(`#${this.id}`).data('department')
+  let supervisor = $(`#${this.id}`).data('supervisor')
+  let data = {"supervisor": supervisor, "department": department}
```
hoerstl: As I mentioned in the python code segment, I'd consider renaming the keys here as such for clarity about what the data holds:
`supervisor` --> `supervisorId`
`department` --> `departmentId`

app/templates/admin/manageDepartments.html
```diff
@@ -62,4 +63,43 @@ <h1>Manage Departments</h1>
...
+        </button>
+        <div class="department-name"></div>
+        
+      </div>
+
+      <div class="modal-body ">
+        <div class="form-group">
+        <label for="supervisorModalSelect">Select a supervisor to add:</label>
+        <div class="dropdown"><select class="selectpicker"
+          name='supervisor'
```
hoerstl: You should consider changing this name to `supervisorId` for accuracy

app/templates/admin/manageDepartments.html
```diff
@@ -62,4 +63,43 @@ <h1>Manage Departments</h1>
...
+                  data-content="{{supervisor.FIRST_NAME}} {{supervisor.LAST_NAME}}<small class='text-muted'> ({{supervisor.ID}})</small>">
+            {{supervisor.FIRST_NAME}} {{supervisor.LAST_NAME}}
+          </option>
+        {% endfor %}
+      </select></div>
+      </div>
+        <div class="changing-content"></div>
+      </div>
+      <div class="modal-footer">
+        <button type="button" id="addSupervisorsToDepartmentSubmit" data-dismiss="modal" class="btn btn-primary">Done</button>
```
hoerstl: Correct me if I'm wrong, but I think that the supervisors are actually added whenever they're accessed via the select and then added to our DB. Therefore, calling this button's id `addSupervisorsToDepartmentSubmit` might be misleading. Consider renaming it to something like `exit`.

app/static/js/manageDepartments.js
andersoncedu: I replaced this with $.when.(function).done()

app/controllers/admin_routes/manage_departments.py
andersoncedu: I don't agree with this, I asked Anderson and we agreed that this would be less descriptive about what data is contained within the variable.

app/logic/search.py
AndersonStettner: A new logic function should be tested.

app/controllers/admin_routes/manage_departments.py
```diff
@@ -29,18 +32,63 @@ def manage_departments():
...
-                                department = department
+                                department = department,
+                                allSupervisors = allSupervisors
                                 )
     except Exception as e:
         print("Error Loading all Departments", e)
         return render_template('errors/500.html'), 500
 
+@admin.route("/admin/manageDepartments/<departmentID>", methods=['GET'])
+def getSupervisorsinDepartment(departmentID):
```
AndersonStettner: in -> In

AndersonStettner (dismissed):

app/logic/search.py
andersoncedu: Added test

app/controllers/admin_routes/manage_departments.py
andersoncedu: Corrected capitalization

app/templates/admin/manageDepartments.html
```diff
@@ -62,4 +63,43 @@ <h1>Manage Departments</h1>
...
+            {{supervisor.FIRST_NAME}} {{supervisor.LAST_NAME}}
+          </option>
+        {% endfor %}
+      </select></div>
+      </div>
+        <div class="changing-content"></div>
+      </div>
+      <div class="modal-footer">
+        <button type="button" id="addSupervisorsToDepartmentExit" data-dismiss="modal" class="btn btn-primary">Done</button>
+        <button type="button" class="btn btn-secondary pull-left closeModal" data-dismiss="modal">Close</button>
```
qasema, line 131: the 'Done' button in the Modal does not confirm or change anything after you add the supervisor. I would suggest that we either remove the button or we allow the change to happen only after the user clicks done

qasema (commented):
there are some test cases failing!  Can you double-check that?

app/templates/admin/manageDepartments.html
andersoncedu: Anderson suggested we leave it there as a way for the user to say "done and save" whereas with close it would be more thought of as "nevermind". Also I think he mentioned it providing symmetry for the UI.

andersoncedu:
> there are some test cases failing! Can you double-check that?

I can't replicate having the tests failing.

andersoncedu:
> there are some test cases failing! Can you double-check that?

I merged development and test are no longer failing.

app/logic/search.py
andersoncedu: Addressed change as requested.

app/static/js/addSupervisorsToDepartment.js
andersoncedu: Addressed this change

azabeli:
image

image


None of the departments are selectable, so we can not even get to where we can attach or remove the supervisors.  I am not sure if that has to do with the department being inactive. But we also have no active department available on the dropdown.

andersoncedu:
> image
> 
> image
> 
> None of the departments are selectable, so we can not even get to where we can attach or remove the supervisors. I am not sure if that has to do with the department being inactive. But we also have no active department available on the dropdown.

This has been addressed

app/controllers/admin_routes/manage_departments.py
```diff
@@ -31,16 +34,60 @@ def manage_departments():
...
+        if not currentUser:                    # Not logged in
+            return render_template('errors/403.html')
+        if not currentUser.isLaborAdmin:       # Not an admin
+            if currentUser.student: # logged in as a student
+                return redirect('/laborHistory/' + currentUser.student.ID)
+            elif currentUser.supervisor:
+                return render_template('errors/403.html'), 403
+        
+        supervisors = getSupervisorsForDepartment(departmentID)
+        departmentID=Department.get_by_id(departmentID)
```
hoerstl: First, the spacing around the '=' sign should be fixed to align with PEP8.
Second, at this point, the variable `departmentID` no longer holds a department ID and should therefore be changed to something like `selectedDepartment`

app/controllers/admin_routes/manage_departments.py
```diff
@@ -31,16 +34,60 @@ def manage_departments():
...
+                                inactiveDepartments = inactiveDepartments,
+                                allSupervisors = allSupervisors
                                 )
     except Exception as e:
         print("Error Loading all Departments", e)
         return render_template('errors/500.html'), 500
 
+@admin.route("/admin/manageDepartments/<departmentID>", methods=['GET'])
+def getSupervisorsInDepartment(departmentID):
+    try:
```
hoerstl: This whole try-except block is unnecessary. When `FLASK_ENV` is set to `staging`, we have our whole application wrapped in a try-except block with the same exception instructions. Therefore, the only time that this block executes is in development where we don't need to be redirected to a pretty page. You should remove this

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -0,0 +1,31 @@
+function addSupervisorToDepartment(supervisorID, departmentID, callback=None) {
```
hoerstl: Rather than using a callback set to none here and checking to see if it's been set later before calling it, you could just define a quick anonymous function that doesn't do anything like:
`callback=() => {}`
then remove the if statement later in the code and always call it

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -0,0 +1,31 @@
...
+    return $.ajax({
+      method: "POST",
+      url: `/supervisorPortal/addUserToDept`,
+      data: {"supervisorID": supervisorID, "departmentID": departmentID},
+      success: function(response) {
+        if (response == "True") {
+          msgFlash("Supervisor has been added to department.", 'success')
```
hoerstl: You should probably double quote 'success' to match the strings around it.

app/static/js/manageDepartments.js
```diff
@@ -26,8 +27,87 @@ $(document).ready( function(){
...
+      attachModalToDepartment()
+    })
+    $('#manageDepartmentSupervisorModal').on('hidden.bs.modal', function() {
+      clearDropdowns()
+    })
 });
 
+
+function attachModalToDepartment() {
+    $('.supervisorDeptModal').off('click')
```
hoerstl: The way you're using this class name here is super confusing. Intuitively, you'd think that the actual modal itself would have the class `supervisorDeptModal` but what really has this class, isn't the modal itself, it's the links that open up the modal. I was sitting here wondering why you were setting an on click for the modal. You should definitely change the name of this class where it appears here on these two lines and then in the HTML. It'd be better as something like `departmentName` so it makes it more clear that clicking on the department name is supposed to update the modal.

hoerstl: Better yet, select for `departmentName` within the tables to be more precise:
`$('.deptTable .departmentName')`

app/static/js/manageDepartments.js
```diff
@@ -26,8 +27,87 @@ $(document).ready( function(){
...
+
+  
+
+function getSupervisorsInDepartment(departmentID) {
+    $.ajax({
+    method: "GET",
+    url: `/admin/manageDepartments/${departmentID}`,
+    success: function(departmentsAndSupervisors) {
+      let currentDepartment=departmentsAndSupervisors[0]
+      let supervisors=departmentsAndSupervisors[1]
```
hoerstl: These two lines can be replaced with the one-line unpacking syntax:
`let [currentDepartment, supervisors] = departmentAndSupervisors`

hoerstl: and make sure you put spaces around the `=` sign

app/static/js/manageDepartments.js
```diff
@@ -26,8 +27,87 @@ $(document).ready( function(){
...
+      
+      let supervisorContent = '<div class="changing-content">'
+      for (let i=0; i<supervisors.length; i++) {
+        let supervisorFirstName= supervisors[i]['preferred_name'] ? supervisors[i]['preferred_name'] : supervisors[i]['legal_name']
```
hoerstl: space before your equals signs

app/static/js/manageDepartments.js
```diff
@@ -26,8 +27,87 @@ $(document).ready( function(){
...
+$("#supervisorModalSelect").on('change', function() {
+  let supervisorID = $('#supervisorModalSelect :selected').val()
+  let departmentID = $('#departmentModalSelect').data('department-id')
+  addSupervisorToDepartment(supervisorID, departmentID, ()=>getSupervisorsInDepartment(departmentID))
+  
+})
+
+  
+
+function getSupervisorsInDepartment(departmentID) {
```
hoerstl: This function's name is quite misleading. Rather than being a getter, which would return the supervisors for a particular department, this function's job is to update the modal with the relevant supervisors for the selected department so it really should be named something like 
`showSupervisorsInDepartment`
or
`updateSupervisorsFromDepartment`

app/static/js/manageDepartments.js
```diff
@@ -26,8 +27,87 @@ $(document).ready( function(){
...
+      $('#manageDepartmentSupervisorModal').modal('show')
+      $('.removeSupervisorFromDepartment').on('click', removeSupervisorFromDepartment)
+    }
+    })
+  }
+
+function removeSupervisorFromDepartment () {
+  let department = $(`#${this.id}`).data('department')
+  let supervisor = $(`#${this.id}`).data('supervisor')
+  let data = {"supervisorID": supervisor, "departmentID": department}
```
hoerstl: You may consider naming these vars `departmentID` and `supervisorID` to match the assignment to the data variable

app/static/js/manageDepartments.js
```diff
@@ -26,8 +27,87 @@ $(document).ready( function(){
...
+          msgFlash("Supervisor has been removed from department.", 'success')
+          getSupervisorsInDepartment(department)
+        } else {
+          msgFlash("Supervisor is not a member of this department.", "warning")
+        }
+      },
+      error: function() {
+        msgFlash("Failed to remove supervisor, please try again.", "fail")
+      },
+    })
```
hoerstl: Watch the indentation here. From this line up to `error: function() {`, we should decrease by one indentation level. The error block should be in line with the success block.

app/templates/admin/manageDepartments.html
```diff
@@ -57,7 +58,7 @@ <h1>Manage Departments</h1>
               <tbody>
                 {% for department in activeDepartments %}
                 <tr>
-                  <td scope="row">{{department.DEPT_NAME}} <small>({{department.ORG}}, {{department.ACCOUNT}})</small></td>
+                  <td scope="row"><a href="#" class="supervisorDeptModal" id="{{department.departmentID}}">{{department.DEPT_NAME}}</a><small>({{department.ORG}}, {{department.ACCOUNT}})</small></td>
```
hoerstl: This is a little more of a personal preference but currently, I dislike how when you click a link it changes the URL at the top of the page to be `/admin/manageDepartments#` instead of leaving it the same. Consider changing the `href=` to `javascript:{}` which doesn't do anything to change the url.

app/templates/admin/manageDepartments.html
```diff
@@ -82,7 +83,7 @@ <h1>Manage Departments</h1>
               <tbody>
                 {% for department in inactiveDepartments %}
                 <tr>
-                  <td scope="row">{{department.DEPT_NAME}} <small>({{department.ORG}}, {{department.ACCOUNT}})</small></td>
+                    <td scope="row"><a href="#" class="supervisorDeptModal" id="{{department.departmentID}}">{{department.DEPT_NAME}}</a><small>({{department.ORG}}, {{department.ACCOUNT}})</small></td>
```
hoerstl: See personal preference comment for the `activeDepartments` table...

app/templates/admin/manageDepartments.html
```diff
@@ -93,4 +94,43 @@ <h1>Manage Departments</h1>
...
+      
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+        <div class="department-name"></div>
+        
+      </div>
+
+      <div class="modal-body ">
```
hoerstl: Remove the trailing space character from your class string. I can't belive you left that in there

app/templates/admin/manageDepartments.html
```diff
@@ -93,4 +94,43 @@ <h1>Manage Departments</h1>
...
+        <div class="form-group">
+        <label for="supervisorModalSelect">Select a supervisor to add:</label>
+        <div class="dropdown"><select class="selectpicker"
+          name='supervisorId'
+          id='supervisorModalSelect'
+          data-live-search='true'
+          title="Supervisor">
```
hoerstl: I'd say that for a single tag like the select that you have here we should be consistent with whether we're using single or double quotes for the attributes. I think it's fine if different tags have different styles but the inconsistency within the tag is iffy to me. Up to you

tests/code/test_search.py
```diff
@@ -219,3 +219,18 @@ def test_getDepartmentsForSupervisors():
...
+        SupervisorDepartment.create(supervisor = 'B00763721', department = 1) 
+        SupervisorDepartment.create(supervisor = 'B12361006', department = 1)
+
+        supervisors = getSupervisorsForDepartment(1)
+        assert len(supervisors) == 2
+    
+        transaction.rollback()
+    
+    supervisors = getSupervisorsForDepartment(1)
+    assert len(supervisors) == 0
```
hoerstl: Consider removing this assertion that department 1 has no supervisors and instead creating a new department that you use for your tests. That way, if someone creates a new entry in our test data doesn't break your test.

hoerstl (changes requested):

hoerstl (approved):

AndersonStettner (approved):

PR #434, opened 2024-02-13
344 fix labor history ordering

### LINKED ISSUE #344: Fix Labor history ordering
Labor history is supposed to show LSFs in the reverse order (i.e., the most future term at the top). Summer and Fall are sorted incorrectly. Others may be as well. 

Related: Fall as a term will not be used in the future. Academic Year (AY) should be used moving forward. This may no longer be an issue once that code is in place.

Screen Shot 2022-08-09 at 11 14 27 AM

doucoureb:
Issue:
The problem is in the sorting order, which is intended to be in reverse chronological order, meaning that the most future term should be displayed at the top. However, the sorting for all terms was incorrect and was instead being ordered by date modified which was improper for the form.

Changes:
- Created ordered values for the term code
- Recreated the ordering function inside of the term model" to "Created a method in the Term class to order objects by term
- Created tests 
- Recreated the ordering function inside of the term model
- Made sure that we were passing our formhistories to the term.sort_by_term method

Testing:
- switch to 344-fix-labor-history-ordering
- source setup.sh and ./database/reset_database.sh from-backup
- flask run
- Navigate to the student search  
- The page should display the labor status forms of the selected student from past terms in reverse order (newest to oldest)
- Also, run the test suite

Attached below is how a proper ordering should look like
labor status form ssdt

### DIFF
--- a/app/controllers/main_routes/laborHistory.py
+++ b/app/controllers/main_routes/laborHistory.py
@@ -59,9 +59,7 @@ def laborhistory(id, departmentName=None):
 
                 if len(authorizedForms) == 0:
                     return render_template('errors/403.html'), 403
-        #Beans: The ordering of these authorizedforms is what we're fixing. The plan is to create a function that tacks on an order_by statement to a query
-        # authorizedForms = sorted(list(authorizedForms),key=lambda f:f.reviewedDate if f.reviewedDate else f.createdDate, reverse=True)
-        #Beans: The following is the tentative order_by statement we'll be putting into a function
+        
         
         authorizedForms = Term.order_by_term(list(authorizedForms.objects()), reverse=True)
 
--- a/app/controllers/main_routes/laborHistory.py
+++ b/app/controllers/main_routes/laborHistory.py
@@ -33,7 +33,7 @@ def laborhistory(id, departmentName=None):
         if not currentUser:                    # Not logged in
             return render_template('errors/403.html'), 403
         student = getOrCreateStudentRecord(bnumber=id)
-        studentForms = (FormHistory.select(FormHistory, LaborStatusForm.termCode.alias('termCode'))
+        studentForms = (FormHistory.select(FormHistory, LaborStatusForm.termCode, LaborStatusForm.jobType)
                                    .join_from(FormHistory, LaborStatusForm)
                                    .join_from(FormHistory, HistoryType)
                                    .where(FormHistory.formID.studentSupervisee == student, 
@@ -45,16 +45,16 @@ def laborhistory(id, departmentName=None):
                 if currentUser.student.ID != id:
                     return redirect('/laborHistory/' + currentUser.student.ID)
             elif currentUser.supervisor and not currentUser.student:
-                supervisorForms = FormHistory.select(FormHistory, LaborStatusForm.termCode.alias('termCode')) \
+                supervisorForms = FormHistory.select(FormHistory, LaborStatusForm.termCode, LaborStatusForm.jobType) \
                                   .join_from(FormHistory, LaborStatusForm) \
                                   .where((FormHistory.formID.supervisor == currentUser.supervisor.ID) | (FormHistory.createdBy == currentUser)) \
                                   .distinct()
-                deptForms = FormHistory.select(FormHistory, LaborStatusForm.termCode.alias('termCode')) \
+                deptForms = FormHistory.select(FormHistory, LaborStatusForm.termCode, LaborStatusForm.jobType) \
                                   .join(LaborStatusForm) \
                                   .join( Department) \
                                   .where(FormHistory.formID.department.DEPT_NAME == currentUser.supervisor.DEPT_NAME) \
                                   .distinct()
-                authorizedForms = studentForms.intersect(supervisorForms.union(deptForms))
+                authorizedForms = studentForms.intersect(supervisorForms.union(deptForms)).order_by(LaborStatusForm.jobType)
 
 
                 if len(authorizedForms) == 0:
@@ -62,7 +62,6 @@ def laborhistory(id, departmentName=None):
         #Beans: The ordering of these authorizedforms is what we're fixing. The plan is to create a function that tacks on an order_by statement to a query
         # authorizedForms = sorted(list(authorizedForms),key=lambda f:f.reviewedDate if f.reviewedDate else f.createdDate, reverse=True)
         #Beans: The following is the tentative order_by statement we'll be putting into a function
-        
         
         authorizedForms = Term.order_by_term(list(authorizedForms.objects()))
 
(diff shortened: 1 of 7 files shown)


app/controllers/main_routes/laborHistory.py
```diff
@@ -69,6 +75,7 @@ def laborhistory(id, departmentName = None):
 
     except Exception as e:
         print("Error Loading Student Labor History", e)
+        raise e
```
bledsoef: Do not raise in exception

andersoncedu:
This appears to work, however I'm unsure why you didn't just order by start date, or how the term codes are different from the start dates.

app/controllers/main_routes/laborHistory.py
```diff
@@ -27,36 +27,42 @@
...
             elif currentUser.supervisor and not currentUser.student:
-                supervisorForms = FormHistory.select() \
+                supervisorForms = FormHistory.select(FormHistory, LaborStatusForm.termCode, LaborStatusForm.jobType) \
                                   .join_from(FormHistory, LaborStatusForm) \
                                   .where((FormHistory.formID.supervisor == currentUser.supervisor.ID) | (FormHistory.createdBy == currentUser)) \
                                   .distinct()
-                deptForms = FormHistory.select() \
+                deptForms = FormHistory.select(FormHistory, LaborStatusForm.termCode, LaborStatusForm.jobType) \
                                   .join(LaborStatusForm) \
                                   .join( Department) \
```
stevensonmichel: There is an unnecessary space in that line of code (before Department).

app/controllers/main_routes/laborHistory.py
```diff
@@ -27,36 +27,42 @@
...
 
                 if len(authorizedForms) == 0:
                     return render_template('errors/403.html'), 403
-        authorizedForms = sorted(authorizedForms,key=lambda f:f.reviewedDate if f.reviewedDate else f.createdDate, reverse=True)
+        
+        
+        authorizedForms = Term.order_by_term(list(authorizedForms.objects()), reverse=True)
+
         laborStatusFormList = ','.join([str(form.formID.laborStatusFormID) for form in studentForms])
         return render_template( 'main/formHistory.html',
```
stevensonmichel: There is an unnecessary line here as well before 'main...'.

tests/code/test_models.py
```diff
@@ -22,3 +26,54 @@ def test_user_model():
...
     assert both_user.email == "bryantal@berea.edu"
+
+
+@pytest.mark.integration
+def test_term_model():
+    def createLSFandFormHistoryObj(*, termCode):
+        """
+        Subprocedure to create LSF and FormHistory objects for a particular termCode with dummy data.
+        """
+        Term.get_or_create(termCode= termCode, termName="idk");
```
andersoncedu: Could you be more descriptive than "idk".

app/models/term.py
```diff
@@ -14,3 +15,34 @@ class Term(baseModel):
...
     isMidyearEvaluationOpen = BooleanField(default=False)
+
+
+    @staticmethod
+    def order_by_term(queryResult, *, reverse=False):
+        """
+        Accepts the results of a query where each object has a `termCode` attribute.
+        To bring selected columns from other tables into an objects direct attributes
+        use the .objects() method on the query.
+
```
stevensonmichel, line 26: That sentence starting with "To bring ..." is not quite clear, as well as the instructions on the use of .objects().

app/controllers/main_routes/laborHistory.py
```diff
@@ -27,36 +27,42 @@
...
             return render_template('errors/403.html'), 403
         student = getOrCreateStudentRecord(bnumber=id)
-        studentForms = FormHistory.select().join_from(FormHistory, LaborStatusForm).join_from(FormHistory, HistoryType).where(FormHistory.formID.studentSupervisee == student,
-        FormHistory.historyType.historyTypeName == "Labor Status Form").order_by(FormHistory.formID.startDate.desc())
-        authorizedForms = set(studentForms)
+        studentForms = (FormHistory.select(FormHistory, LaborStatusForm.termCode, LaborStatusForm.jobType)
+                                   .join_from(FormHistory, LaborStatusForm)
+                                   .join_from(FormHistory, HistoryType)
+                                   .where(FormHistory.formID.studentSupervisee == student, 
+                                          FormHistory.historyType.historyTypeName == "Labor Status Form"))
```
andersoncedu, line 43: There are slashes used to separate lines just below this, so either change this to use slashes or change the slashed lines to be wrapped in the parentheses instead.

app/models/term.py
```diff
@@ -14,3 +15,34 @@ class Term(baseModel):
...
+            '00' : 0,
+            '11' : 2,
+            '04' : 3,
+            '01' : 4,
+            '02' : 5,
+            '12' : 6,
+            '05' : 7,
+            '03' : 8,
+            '13' : 9,
+
```
stevensonmichel, line 42: This dictionary maps the last two digits of the term code to numbers. I wonder why you skipped 1 in that case.

tests/code/test_models.py
hoerstl: Addressed with extremely disturbing python syntax.

app/models/term.py
hoerstl: We skipped 1 because all terms that have never been identified before should default to 1 as per line 31 here

app/models/term.py
hoerstl: This comment was just to give people a hint and push them to the peewee documentation. Out of respect for your concern I've cleared up the comment and made it more clear that they should read the official docs.

app/controllers/main_routes/laborHistory.py
hoerstl: Good catch

app/controllers/main_routes/laborHistory.py
hoerstl: Great Catch!

app/controllers/main_routes/laborHistory.py
hoerstl: Good catch, I believe I've resolved the ones below to use parentheses.

app/models/term.py
stevensonmichel: You are pushing people to Peewee. I feel like you did aim specifically for someone.

stevensonmichel:
Lawrence and Bineta, after I search for a student, when I click on Return, it goes to the Supervisor Portal page. The right redirection should be at the Student Search page. I was considering creating a new issue for that, but I believe you both have the skills to fix it, as you are already working on an issue of the same page.

AndersonStettner (approved):

hoerstl:
> Lawrence and Bineta, after I search for a student, when I click on Return, it goes to the Supervisor Portal page. The right redirection should be at the Student Search page. I was considering creating a new issue for that, but I believe you both have the skills to fix it, as you are already working on an issue of the same page.

We've added this functionality as well!

Issue #442, opened 2024-09-05
Creating more filters for Form Search 


WackyWeaver:
## Issue
There are not enough ways to filter between students during an academic year, The only option for time are for active terms or full academic years. Filtering for primaries or secondaries would be helpful when supervisors need to worry allocations. 

## Changes to Make
- Be able to filter by primaries/secondaries
- Hide past ones with toggle (past default).
- Simplify term options to “Breaks”, “Summer”, “AY”
- Turn sorting buttons in column header into a drop down that allows you to sort by a certain field
  - Term
  - Department
  - Supervisor: First Name
  - Supervisor: Last Name
  - Student: First Name
  - Student: Last Name
  - Position
  - Hrs

## Resources
- https://datatables.net/extensions/searchpanes/examples/customFiltering/customOptionConditions.html

WackyWeaver:
![](https://github.trello.services/images/mini-trello-icon.png) [LSF Creating more filters for Form Search #442](https://trello.com/c/TauWmb0n/820-lsf-creating-more-filters-for-form-search-442)

PR #447, opened 2024-09-19
Add additional form search filters

bledsoef:
## Issue
Fixes issue #442 

Right now sorting is not entirely intuitive, and if it is, it is not particularly dynamic. There may be 3 different fields stored in one column (what if I want to sort first name vs last name) but right now all we can do is sort by the text in column. We need to add our own ordering functionality so we can sort by whatever we deem necessary.

## Changes
- Disabled current built in sorting mechanisms because they will now break and are no longer needed
- Added 3 drop downs to select which column you want to sort by, what field in that column you want to sort by, and whether you want it ascending or descending.
- Pass back a string containing the field we want to order by to the backend
- Added in backend logic to sort by particular field instead of one field per column.
- Upgraded the current datatables package to the most up to date one.

## To Test
- Checkout `form_search_filters_442`
- Reset from backup data
- Go to the supervisor portal, experiment around with some queries and try and verify that every field and column work as expected.

### DIFF
--- a/app/static/js/supervisorPortal.js
+++ b/app/static/js/supervisorPortal.js
@@ -23,9 +23,17 @@ $(document).ready(function () {
     if (!isDisabled && $('#fieldPicker').val() == '') {
       msgFlash("Cannot sort without selecting a field.", 'warning')
       return
-    } 
+    }
     runFormSearchQuery()
   })
+
+  if ($('#formSearchTable').is(':hidden')) {
+    $('#columnPicker').selectpicker('hide')
+    $('#fieldPicker').selectpicker('hide')
+    $('#orderPicker').selectpicker('hide')
+    $('#sortByButton').hide()
+  }
+
   $('#addUser').on('click', function () {
     let supervisorID = $('#supervisorModalSelect :selected').val()
     let departmentID = $('#departmentModalSelect :selected').val()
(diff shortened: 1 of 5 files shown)


bledsoef:
@Karina-Agliullova Accidentally ran a linter on the HTML so the diff is pretty large but we really just added the column, field, and order pickers and some styling.

stevensonmichel:
The sorting by Hrs. does not seem to work.

app/static/js/supervisorPortal.js
```diff
@@ -1,184 +1,254 @@
...
+    let supervisorID = $('#supervisorModalSelect :selected').val()
+    let departmentID = $('#departmentModalSelect :selected').val()
 
-      addSupervisorToDepartment(supervisorID, departmentID)
+    addSupervisorToDepartment(supervisorID, departmentID)
   })
   $('#departmentModalSelect').on('change', disableButtonHandler)
   $('#supervisorModalSelect').on('change', disableButtonHandler)
-  $('#clearSelectionsButton').on('click', function(){
+  $('#firstName').on('click', function () {
```
Karina-Agliullova: First name and last name checkboxes are not needed

BrianRamsay (approved):

Issue #441, opened 2024-09-05
Form Search Views


WackyWeaver:
## Problem 
The form search and results cannot be seen together without scrolling. There is no labor history on students outside of their current positions

## Changes to Make
- Have a complex view and a simple view that is more accessible
  - Complex view would contain all of the current information in the table
  - Plain view should resemble earlier versions that contained fewer and more prioritized columns
  - Be able to see both the form search and the results at the time

## Old Supervisor Portal That We Can Use as Simple View
```
<td>
  <a href="/laborHistory/{{form.formID.studentSupervisee.ID}}" value=0>
    <span class="h4">{{form.formID.studentSupervisee.FIRST_NAME}}
                            {{form.formID.studentSupervisee.LAST_NAME}}
                            ({{form.formID.studentSupervisee.ID}})
    </span>
  </a>
  <span class="pushRight h5">{{form.status}}</span>
  <br>
  <span class="pushLeft h6"> {{form.formID.termCode.termName}} - {{form.formID.POSN_TITLE}} - {{form.formID.department.DEPT_NAME}}</span>
</td>
```

WackyWeaver:
![](https://github.trello.services/images/mini-trello-icon.png) [Form Search View and Labor History](https://trello.com/c/zZDNH0FN/821-form-search-view-and-labor-history)

PR #459, opened 2024-10-17
Create simple and advanced views for supervisor portal

WackyWeaver:
## Issue
Fixes issue #441 

We created a new supervisor portal, and that is great, but we took it from a user friendly (albeit a very crappily designed) experience to a big table with a ton of info. We should give them the option to view in a clearer "simpler" view while maintaining all the same functionality and features of the more advanced view.

## Changes
- Created a button in the top right that allows you to toggle between a simple and advanced view.
- Created a new datatable that renders when Switch To Simple View is toggled that limits the information and presents it in a slightly more user friendly way.
- Added new logic in Flask to handle the rendering of a simple view.
- Modified the cookie handling logic so when the data is pulled in from the session cookies it renders the correct view.

## To Test
- Checkout `441simple_view`
- Run `./database/reset_database.sh from-backup` and `export FLASK_ENV=staging`
- `flask run`
- Upon load you should see the supervisor portal.
- Execute a search and verify that it displays properly on the simple view.
- Try sorting and make sure you can only sort by columns and fields that are directly visible to the user.
- Click "Switch to Advanced View" and verify that the data is the same but just displayed differently.
- Try sorting and make sure you can sort by all columns and fields
- Execute a search with advanced view and close the window
- Reload the window and make sure that all data is automatically loaded in properly.

### DIFF
--- a/app/static/js/supervisorPortal.js
+++ b/app/static/js/supervisorPortal.js
@@ -12,7 +12,16 @@ $(document).ready(function () {
 
   $('#formSearchButton').on('click', function () {
     runFormSearchQuery();
-
+  });
+  $('#switchViewButton').on('click', function () {
+    let buttonVal = this.val()
+    if (buttonVal == 'advanced') {
+      this.val('simple')
+      this.html('Switch To Advanced View')
+    } else {
+      
+    }
+    runFormSearchQuery();
   });
   $('#addUserToDept').on('click', function () {
     $("#addSupervisorToDeptModal").modal("show");
@@ -129,6 +138,7 @@ function disableButtonHandler() {
 }
 
 function runFormSearchQuery(button) {
+  let view = $('#switchViewButton').val()
   let termCode, departmentID, supervisorID, studentID;
   let formStatusList = [];
   let formTypeList = [];
@@ -178,6 +188,7 @@ function runFormSearchQuery(button) {
   }
 
   queryDict = {
+    'view': view,
     'termCode': termCode,
     'departmentID': departmentID,
     'supervisorID': supervisorID,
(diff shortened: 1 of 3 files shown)


Karina-Agliullova (approved):
Looks good!

app/static/js/supervisorPortal.js
```diff
@@ -70,8 +89,14 @@ $(document).ready(function () {
   });
   $('#columnPicker').on('change', function () {
     let column = $('#columnPicker :selected').text()
-    let fields = columnFieldMap[column]
+    buttonVal = $("#switchViewButton").val()
+    let fields
+    if (buttonVal == "advanced") {
+      fields = advancedColumnFieldMap[column]
```
Mbeweg: Make this variable declaration and if statement a one-liner using a ternary operator
`let fields = buttonVal == "advanced"? advancedColumnFieldMap[column]:simpleColumnFieldMap[column];`

app/static/js/supervisorPortal.js
```diff
@@ -168,6 +202,9 @@ function runFormSearchQuery(button) {
   }
 
   queryDict = {
+    // this may seem counterintuitive but since the value
+    // switches on button press it is technically the opposite of the current view
```
Mbeweg: These comments are out of date. The value of the switchViewButton **is equal to** the current view in the most recent commit. Should probably remove or update.

Mbeweg (commented):
Changes Requested

app/static/js/supervisorPortal.js
```diff
@@ -182,14 +214,83 @@ function runFormSearchQuery(button) {
...
+    $('#switchViewButton').html('Switch To Simple View')
+    $('#columnPicker').empty();
+    let columns = Object.keys(advancedColumnFieldMap)
+    columns.forEach((column) => {
+      var option = $('<option>', {
+        value: advancedColumnFieldMap[column][0][1],
+        text: column
+      });
+      $('#columnPicker').append(option)
+    })
```
BrianRamsay: Let's pull out 240-248 into a resetColumnPicker function (or some other name), as it is basically identical to the code in the other branch (228-236). That will make each side of the conditional look a lot simpler, and we won't repeat the code.

BrianRamsay (changes requested):
Just one small change

BrianRamsay (approved):

PR #462, opened 2024-11-05
Add permission for admin students for New Labor Status Form

stevensonmichel:
## Issue: 
Previously, students who are also Labor Admins did not have access to the New Labor Status Form page. That was considered as a bug given that these students have the permissions to add submit labor status forms for students. 

## Solution: 
Through exploration, we discovered that in the backend such a logic to allow labor admin students to see the New Labor Status Form was already implemented. However, the sidebar was not displaying the New Labor Status Form for these students. So, we made changes there to reflect the privileges of accessibility for these students. 

## Test:
Assuming you're in the LSF directory
- source setup.sh
- database/reset_database.sh from-backup
- change the login username into a student who is a Labor Admin (ex. floresd)
- export FLASK_ENV=staging
- flask run
Now open the application and verify if New Labor Status Form is present as an option in the sidebar. 

Fixes Issue #1321
https://github.com/BCStudentSoftwareDevTeam/celts/issues/1321

### DIFF
--- a/app/templates/sidebar.html
+++ b/app/templates/sidebar.html
@@ -24,13 +24,17 @@
                   </div>
               </a>
           </div>
+      {% endif %}
+      {% if currentUser.supervisor or (currentUser.student and currentUser.isLaborAdmin) %}
           <div class="panel panel-default">
             <a class="adminTest" href="/laborstatusform">
-                <div {% if request.path =="/laborstatusform" %} class="panel-heading active"{% endif %} class="panel-heading" tabindex="5">
+                <div {% if request.path == "/laborstatusform" %} class="panel-heading active"{% endif %} class="panel-heading" tabindex="5">
                       <h4>New Labor Status Form</h4>
                   </div>
               </a>
           </div>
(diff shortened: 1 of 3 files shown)


app/templates/sidebar.html
```diff
@@ -39,6 +39,14 @@ <h4>Student Search</h4>
               </a>
           </div>
         </div>
+      {% elif currentUser.student and currentUser.isLaborAdmin %}
```
bledsoef: Could we potentially add this conditional to the other New Labor Status Form instead of duplicating all this code?

bledsoef (changes requested):
Functionality looks alright besides some tracy problems which are unrelated. Just might try to prevent duplicating code. Nice job!

app/templates/sidebar.html
hoerstl: Yessir. I've replaced the single if statement above with three so the sidebar components remain in order and removed the duplicate code.

bledsoef (approved):
LGTM

PR #479, opened 2025-04-18
Fix for LSF Downloads + QOL Improvements

### LINKED ISSUE #376: Form Search download doesn't match table
In many cases, the form search download doesn't match what is in the datatable. It should always be identical (in terms of # of results).

ojmakinde:
Fixes #349 
Fixes #376

Downloads do not work.

What we did:
- Fixed the downloads.
- Replaced global variables (which are not thread-safe and lead to null download errors on page refresh) and replaced them with cookies for better consistency.
- Refactored the CSVMaker module and removed unnecessary DB queries + logic.
- Updated downloads to use better naming conventions, instead of the default "LaborStudents.csv"

To test:
- Pull the branch
- Test Supervisor Portal downloads as a Labor Admin (heggens is a Labor Admin)
- Test downloads with midyear and final evaluations
- Switch to Administration > Pending Forms
- Download as Labor Admin and Financial Admin. Any downloads with Financial Admin should have additional spreadsheet columns.

A good follow-up to this PR should be #484

### DIFF
--- a/app/controllers/admin_routes/allPendingForms.py
+++ b/app/controllers/admin_routes/allPendingForms.py
@@ -208,17 +208,21 @@ def checkAdjustment(allForms):
 
 @admin.route('/admin/pendingForms/download', methods=['POST'])
 def downloadAllPendingForms():
+    allPendingFormsSelectObject = request.form.get('formList')
+    print(allPendingFormsSelectObject, type(allPendingFormsSelectObject), "the-type")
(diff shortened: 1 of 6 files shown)


BrianRamsay (approved):
I'm merging so we can get most download functionality on production.

Comments below:
- The Download button on Labor History is still broken
- Add a block comment explaining cookie usage where they are set
- Can we fix the ordering of the csv records so that they are in the same order as the report? At least the report default ordering.

PR #490, opened 2025-06-26
Imported load_config function from loadConfig.py to contributors.py

### LINKED ISSUE #482: SSDT Footer Link Does Not Work
On LSF, clicking the SSDT link on the page footer does not work.

Before clicking:

Image

After clicking:

Image

mathew2005:
## Issue Description

Fixes #482 
- SSDT Footer Link Does Not Work

## Changes
- Fixed contributors page by importing load_config function from `loadConfig.py` to `contributors.py`.
- Also removed duplicate code `getAbsolutePath` by deleting `absolute_path.py`

## Testing
- Login as any user (should be logged in by default)
- Go to the page's footer and click Student Software Development Team link.
image
- Should redirect you and show you a table of all the contributors stored in `contributors.yaml`
image


app/controllers/main_routes/contributors.py
```diff
@@ -6,6 +6,7 @@
 from app.models.Tracy.studata import *
 from app import *
 from app.login_manager import *
+from app.config.loadConfig import *
```
BrianRamsay: Even though we are using `*` everywhere else, it's generally not a good idea. Instead of importing everything, just import the symbols you need. In this case, `getAbsolutePath`.

BrianRamsay (changes requested):

app/controllers/main_routes/contributors.py
mathew2005: @BrianRamsay I will only need the `load_config` function, as the `getAbsolutePath` is defined and used only on localConfig.py. I was able to change it so that it only imports `load_config`.

BrianRamsay (approved):

Issue #451, opened 2024-10-11
Logout page doesn’t work


Karina-Agliullova:
Logout page doesn’t log you out and only redirects to basic HTML template (At least on student side). Fix the logout issue

PR #492, opened 2025-07-03
Logout issue 451

RueHaile:
Resolved issue: [451_ LSF](https://github.com/BCStudentSoftwareDevTeam/lsf/issues/451)

_Fixed:_

- The log out button on LSF student side has been fixed so that they will be logged out successfully and also  communicated about it by removing the remove error message and showing a shibboleth logged out message on a new web page.

_Issue:_

 -Logout page doesn’t log you out and only redirects to basic HTML template (On student side). 
- It gave a maximum recursion depth reached error, on the side of development and an unexpected error occurred on the production side when the user attempts to log out. 
-The user was not informed that they have successfully checked out only that there has been:
“Unexpected error has occurred, an administrator has been notified and sorry for the problem page”.

**_Testing the issue :Development error_** 
image

**_Testing the issue: production/published side_** 
image 
 
**_Currently tested to output/ resolved :_**  
image

**_When the user is logged out, they are redirected to the Duo login page:-_**
image
**_If the user presses back on the web browser, a stale request will be rendered._** 
image

 

 
Test: 
•	Normally, students the way that we can get the labor form website if we are provided with a link from a supervisor.
•	Check out branch:   logout_issue_45.
•	Once on the page, there is a sidebar with a logout button on it. 
•	Click the log out and followed by a “logged-out operation is complete page.
•	The user has successfully logged out; they can exit the page.

### DIFF
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -17,12 +17,12 @@ from app.controllers.admin_routes.allPendingForms import checkAdjustment
 from app.controllers.main_routes import main_bp
 from app.controllers.main_routes.download import CSVMaker
 from app.logic.search import getDepartmentsForSupervisor
-from app.login_manager import require_login, logout
-
+from app.login_manager import require_login, logout as performlogout
 
 @main_bp.route('/logout', methods=['GET'])
-def logout():
-    return redirect(logout())
+def handlelogout():  
+    performlogout() 
+    return redirect('https://www.berea.edu/labor-program-office')
 
 @main_bp.route('/', methods=['GET', 'POST'])
 def supervisorPortal():
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -21,7 +21,8 @@ from app.login_manager import require_login, logout as performlogout
 
 @main_bp.route('/logout', methods=['GET'])
 def handlelogout():  
-    performlogout() 
+    performlogout()
+    return redirect("https://secure.berea.edu/Shibboleth.sso/Logout?return=https://lsf.berea.edu/logged_out") 
    
 @main_bp.route('/', methods=['GET', 'POST'])
 def supervisorPortal():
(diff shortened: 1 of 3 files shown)


bakobagassas:
Everything looks fine

adjagbodjouf:
After logging out, the successful logout page loads, indicating that the user has been logged out. But after this page shows up, you can click the back arrow everything still works fine, like the user still hasn't logged out. Even after refreshing, everything appears to still be working when it shouldn't. This may just be because this is all running locally, but if not, then it may be something you need to fix.



**Responses:**
- The back arrow won't be rendering the web page anymore without the user logging in.
-  The back button will not work when they are logged out.
**_This issue has been resolved._**

adjagbodjouf:
This would look better centered on the page instead of being left all the way at the top.

image

The text might also look better centered in the box.

RueHaile:
> This would look better centered on the page instead of being left all the way at the top.
> 
> image
> 
> The text might also look better centered in the box.

**Response:** 
- This is no longer an issue because we are rendering a different page.
_**This issue has been resolved.**_

BesherKitaz:
I got this error when trying to open the site on the test database when I try to open the website as a student (I set the user to 'kitazb' in secret_config.yaml, which is a student username):


image

When I set the username back to "heggens" to view the supervisor portal:

image

And then when I clicked "back," it took me to my profile in lsf production site:

image

RESPONSE: 
Clearing the cache should resolve the issue.

BrianRamsay (changes requested):
This fixes the recursion error, but doesn't actually log them out. We need to figure out what we need to configure so that going to `https://secure.berea.edu/Shibboleth.sso/Logout` actually ends their duo/shibboleth session.

If we can't figure that part out, we should just remove the Logout link altogether.

BrianRamsay:
I implemented just the logout fix in commit 5d2fca22. Had to find the right url redirect and make a config change in `/etc/shibboleth/shibboleth2.xml` in production.

PR #493, opened 2025-07-03
resubmit button on approved forms

### LINKED ISSUE #419: Create a Banner resubmit button
Resubmit button so that something that should have pushed to banner can be hit so that it goes into Banner.

This button should only be for admins and it should be on approved forms.

jjijon7000:
fixes #419 

Description:

Resubmit button so that something that should have pushed to banner can be hit so that it goes into Banner.
This button should only be for admins and it should be on approved forms.

Changes:

- added a button to the studenthistorymodal html, added the ajax in the laborhistory js file, and added the route in allpendingforms.py
- fixed sub-issue of the forever loading icon gif

Testing:

*Before*
image
^was stuck like this forever basically

*After*
image

- run flask
- reset the database to *from-backup* instead of *test*
- select a term and search
- select a student
- click on any of their approved forms and see the changes


PuskarC:
lsf

The current implementation causes the Supervisor Portal page to load without displaying any data or interactive elements. Also, the page appears unresponsive as no results are shown under the "Form Search" section

app/controllers/admin_routes/allPendingForms.py
```diff
@@ -816,3 +816,43 @@ def getNotesCounter():
     except Exception as e:
         print("Error selecting admin notes:", e)
         return jsonify({"Success": False}),500
+
+@admin.route('/admin/addToBanner/<form_id>', methods=['POST'])
+
+
+def submitToBanner(form_id):
```
BrianRamsay: We already have a function that submits to banner, right? It happens whenever a form is approved by admins. We shouldn't be re-inventing that solution.

app/controllers/admin_routes/allPendingForms.py
```diff
@@ -816,3 +816,43 @@ def getNotesCounter():
...
+            conn = Banner()
+            save_status = conn.insert(labor_form)
+
+            if save_status:
+                return jsonify({"success": True})
+
+            else:
+                return jsonify({"success": False}), 500
+        else:
+            return jsonify({"success": False})
```
BrianRamsay: Don't use success: True and False. Use the status codes to determine success or not.

I suspect you also need a try/except around the banner save attempt, but really you should be re-using the code that already does this.

app/controllers/admin_routes/allPendingForms.py
```diff
@@ -816,3 +816,43 @@ def getNotesCounter():
     except Exception as e:
         print("Error selecting admin notes:", e)
         return jsonify({"Success": False}),500
+
+@admin.route('/admin/addToBanner/<form_id>', methods=['POST'])
+
+
```
BrianRamsay: Remove the space between the route and the method, please

BrianRamsay (changes requested):

app/templates/snips/studentHistoryModal.html
```diff
@@ -213,6 +213,7 @@
       <div class = "col-sm-{{12/buttonState.num_buttons}}">
 
       <a href="#"><button class="btn btn-primary" id="emailLabor">Email Labor</button></a>
+      <button type="button" class="btn btn-primary" data-dismiss="modal" onclick="submitToBanner({{statusForm.laborStatusFormID}})">Resubmit Form</button>
```
BrianRamsay, line 216: Do you see how all of the other buttons use the buttonState object to determine if the button should display? This is necessary because it doesn't make sense to have a Resubmit button on every form, and certainly not for every user. Students should not be able to submit their forms to banner, for example.

BrianRamsay (changes requested):

app/templates/snips/studentHistoryModal.html
BrianRamsay: Has not been resolved yet

BrianRamsay:
Button state changes were never addressed. Re-implemented in commit 87211c28.

PR #494, opened 2025-07-03
Organizingtest - Reorganzing LSF Directory Structure

BesherKitaz:
## Issue Description

Fixes #487
We need to reorganize the files within LSF.  All Logic functions  within the controller section and need to be moved them into the logic section. New files need to be created in logic to match routers and import statements need to be handled properly.

## Changes

- Removed all functions that didn't contain routes from most controller files to logic
- Created five new files in logic: alterLSF in logic, departmentPositions, statusFormFunctions, studentsFromTracy, and supervisorAdjustment.py.
- The new files we created contain the moved functions from controller. (There have been no created or split up functions only moved)
- We also moved downloads.py to logic since it didn’t have any routes.
- Adjusted import statements accordingly to match the new structure

## Testing

- This is a unique case since there are no changes in the functionality or UI; only the project structure. It is best to check that the website is running properly, no import errors, and maybe test the functions whose location changed.


petrosyantsa:
I think the new structure of the program looks good, we runed the code and everything seems to be working.

app/logic/adminManagement.py
```diff
@@ -0,0 +1,92 @@
...
+            supervisors.append(supervisor)
+    for sup in supervisors:
+        username = sup.EMAIL.split('@', 1)
+        userList.append({'username': username[0],
+                        'firstName': sup.FIRST_NAME,
+                        'lastName': sup.LAST_NAME,
+                        'type': 'Supervisor'})
+    return userList
+
+def getUser(selectpickerID):
```
BrianRamsay, line 52: `getUser` doesn't seem inherently admin-related

app/logic/alterLSF.py
```diff
@@ -0,0 +1,140 @@
...
+                                           status       = status.statusName)
+        
+        if fieldName == "weeklyHours":
+            newWeeklyHours = int(fieldsChanged[fieldName]['newValue'])
+            createOverloadForm(newWeeklyHours, lsf, currentUser, adjustedforms.adjustedFormID, adjustedFormHistory,host=host)
+
+        return adjustedFormHistory.formHistoryID
+    
+
+def createOverloadForm(newWeeklyHours, lsf, currentUser, adjustedForm=None,  formHistories=None, host=None):
```
BrianRamsay: This seems like a better fit for `statusFormFunctions.py`

app/logic/studentsFromTracy.py
```diff
@@ -0,0 +1,9 @@
+from app.logic.tracy import Tracy
+from app.logic.tracy import InvalidQueryException
+from peewee import DoesNotExist
+from app.models.student import Student
+from app.models.user import User
+
+
+
+
```
BrianRamsay: I don't think this file is supposed to exist

app/logic/supervisorAdjustment.py
```diff
@@ -0,0 +1,37 @@
...
+            newPosition = Tracy().getPositionFromCode(newPositionCode)
+            # temporarily storing the position code and wls in new value, and position name in old value
+            # because we want to show these information in the hmtl template.
+            allForms.adjustedForm.newValue = newPosition.POSN_CODE +" (" + newPosition.WLS+")"
+            allForms.adjustedForm.oldValue = newPosition.POSN_TITLE
+
+        if allForms.adjustedForm.fieldAdjusted == "department":
+            newDepartment = Department.get(Department.ORG==allForms.adjustedForm.newValue)
+            allForms.adjustedForm.newValue = newDepartment.DEPT_NAME
+            allForms.adjustedForm.oldValue = newDepartment.ORG + "-" + newDepartment.ACCOUNT
```
BrianRamsay: It seems like this function is cut off halfway. Is this the whole thing? Either way, I'm not sure it makes sense to have a separate file for one function.

app/logic/user.py
```diff
@@ -0,0 +1 @@
+# To be removed
```
BrianRamsay: Needs to be removed

app/logic/utils.py
```diff
@@ -15,3 +17,16 @@ def makeThirdPartyLink(recipient, host, formHistoryId):
 
 def setReferrerPath():
     session['referrerPath'] = urlparse(request.referrer).path or ''
+
+
+def adminFlashMessage(user, action, adminType):
```
BrianRamsay, line 22: You have this method defined in `adminManagement.py` already

app/controllers/admin_routes/adminManagement.py
```diff
@@ -1,14 +1,16 @@
...
 from app.controllers.admin_routes import admin
 from flask import request
 from app.login_manager import require_login
 from flask import Flask, redirect, url_for, flash, jsonify
 from app.models.supervisor import Supervisor
 from app.models.student import Student
-from app.logic.tracy import Tracy
-from app.logic.userInsertFunctions import createUser, createSupervisorFromTracy, createStudentFromTracy
+from app.logic.tracy import Tracy, createStudentFromTracy
+from app.logic.userInsertFunctions import createSupervisorFromTracy, createUser
```
BrianRamsay: the tracy logic file has `createStudentFromTracy` but not `createSupervisorFromTracy`? Seems like those two should be in the same file. If they already are, the imports are incorrect here.

BrianRamsay (changes requested):
This is going to be a nice change!

app/logic/supervisorAdjustment.py
BesherKitaz: Potentially yes, it maybe cut off, but this is how it looks like in development. So even if it is cut off, we can check if need to create a new issue for it. 

For the function location, we moved it to the logic/addPendingForms.py. the reason we chose this path is because it was in admin_routes/allPendingForms.py. We just moved to the logic section cuz it does more logic and it is not a route.

app/logic/user.py
BesherKitaz: Removed.

app/logic/studentsFromTracy.py
BesherKitaz: Removed.

app/logic/alterLSF.py
BesherKitaz: Moved to statusFormFunctions.py as asked.

app/logic/utils.py
BesherKitaz: We removed it from adminManagement when we created the utils.py file.

app/logic/adminManagement.py
BesherKitaz: Yes that is true. However, the function is only used in this file and has no imports at all. So it makes to keep in here so that it is easy to access when working with adminManagement. Also, I am not sure if it fits into any other files, so it makes sense to keep it in the one where it is only used to avoid unnecessary import statements.

app/controllers/admin_routes/adminManagement.py
BesherKitaz: Done. Moved to UserInsertFunctions.py.

app/logic/utils.py
BrianRamsay: It's still there in the PR

app/controllers/main_routes/main_routes.py
```diff
@@ -1,24 +1,17 @@
...
+from peewee import fn
 from app.models.department import Department
 from app.models.supervisor import Supervisor
 from app.models.supervisorDepartment import SupervisorDepartment
 from app.models.student import Student
 from app.models.laborStatusForm import LaborStatusForm
 from app.models.formHistory import FormHistory
-from app.models.user import User
-from app.models.studentLaborEvaluation import StudentLaborEvaluation
 from app.controllers.admin_routes.allPendingForms import checkAdjustment
```
BrianRamsay, line 9: It's no longer in this file.

app/logic/supervisorAdjustment.py
BrianRamsay: If the function exists in `allPendingForms.py` now, we should get rid of this file.

BrianRamsay (changes requested):

app/controllers/main_routes/main_routes.py
BesherKitaz: did as you asked for the first comment

BesherKitaz: And the 3rd comment too.

PR #495, opened 2025-07-10
Labor release form

### LINKED ISSUE #458: Labor Release forms should not be a surprise to the Labor Office staff
At LPC, it was indicated that Labor Release forms sometimes are not communicated to the Labor Office prior to the form being put into the system. They then sit there for weeks before getting addressed. 

I propose we add a field to the Release form which indicates which Labor Office staff member was spoken to about the release form, and make it a required field. There should also be language added to the preamble about the expectation that they speak with Labor first. 

I would also propose that an email goes directly to the selected staff member indicating the form was put in.

Meatchema:
## Issue Description
Fixes #458
- We had to figure out how to add a required name selection to the labor release form so that users would add who they talked to within the labor office regarding the form. This involved also adding a column within the sql database so that the information is also taken into the system for record. The column is for the person they contacted before filling out a release form.

Note: a second part of the problem, which should be handled in a different issue and pull request, is sending an email to the contact person (the field we just added) that the user selects when filling out the labor release form so that this person is notified and can start looking for a new labor position for the released student.
-
## Changes
- Added an Alter Table in prod-backup.sql to add a column within the database for the contactPerson_ID.
- Added .space and .studentspace to the labor release form css file to fix its formatting
- Added a drop down selector in the html file to add a required labor office staff entry section.

## Testing
-Switch the FLASK_ENV environment to development
- Reset database from the backup database (from-backup)
- Switch back the FLASK_ENV to staging
- Go to secret config: Make sure the user is a laborAdmin (heggens, for example) and make sure the Tracy password is in place
-Click on the New Labor Status Form Section from the navigation panel on the left and create a labor status form so we can use it for testing
-Go to slack (laborstatusform channel) to view the email that is meant to be sent to the student and accept the form
-Then go to administration and click pending forms, checkbox the most recent one that you created (at the top), scroll down to the bottom and scroll down and click "Approve selected" and approve the form
-Then click on the supervisor portal and enter the filters pertaining to your form, and then select that form.
-Click on your form and then click the release button then click submit to see the release form page, submit after filling out with data
- Observe the new change (the new field Contact Person field), and then check if the new field is added to the database in the created record (the field contactperson_id)

### DIFF
--- a/app/controllers/main_routes/laborReleaseForm.py
+++ b/app/controllers/main_routes/laborReleaseForm.py
@@ -26,6 +26,7 @@ def laborReleaseForm(laborStatusKey):
             return redirect('/laborHistory/' + currentUser.student.ID)
 
     forms = LaborStatusForm.select().distinct().where(LaborStatusForm.laborStatusFormID == laborStatusKey)
+    laborAdmins = User.select().where(User.isLaborAdmin == 1)
 
     if(request.method == 'POST'):
         try:
@@ -47,8 +48,10 @@ def laborReleaseForm(laborStatusKey):
             releaseDate = datetime.strptime(datepickerDate, "%m/%d/%Y").strftime("%Y-%m-%d")
             releaseReason = request.form.get("notes")
             releaseCondition = request.form.get("condition")
+            releaseContactUsername = request.form.get("contactPerson")
+            releaseContactPerson = User.select().where(User.username == releaseContactUsername).get()
 
-            createLaborReleaseForm(currentUser, laborStatusForiegnKey, releaseDate, releaseCondition, releaseReason)
+            createLaborReleaseForm(currentUser, laborStatusForiegnKey, releaseDate, releaseCondition, releaseReason, releaseContactPerson)
 
             email = emailHandler(formHistoryID.formHistoryID)
             email.laborReleaseFormSubmitted()
@@ -66,14 +69,16 @@ def laborReleaseForm(laborStatusKey):
             return redirect(url_for("main.supervisorPortal"))
     return render_template('main/laborReleaseForm.html',
 				            title=('Labor Release Form'),
-                            forms = forms
+                            forms = forms,
+                            laborAdmins=laborAdmins
                           )
 
-def createLaborReleaseForm(currentUser, laborStatusForiegnKey, releaseDate, releaseCondition, releaseReason, status="Pending", reviewedDate=None, reviewedBy=None):
+def createLaborReleaseForm(currentUser, laborStatusForiegnKey, releaseDate, releaseCondition,  releaseReason, releaseContactPerson, status="Pending", reviewedDate=None, reviewedBy=None):
     newLaborReleaseForm = LaborReleaseForm.create(
                                 conditionAtRelease = releaseCondition,
                                 releaseDate = releaseDate,
-                                reasonForRelease = releaseReason
+                                reasonForRelease = releaseReason,
+                                contactPerson= releaseContactPerson
                                 )
     historytype = HistoryType.get(HistoryType.historyTypeName == "Labor Release Form")
     status = Status.get(Status.statusName == status)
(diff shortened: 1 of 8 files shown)


app/controllers/main_routes/laborReleaseForm.py
```diff
@@ -47,8 +48,10 @@ def laborReleaseForm(laborStatusKey):
             releaseDate = datetime.strptime(datepickerDate, "%m/%d/%Y").strftime("%Y-%m-%d")
             releaseReason = request.form.get("notes")
             releaseCondition = request.form.get("condition")
+            releaseContactUsername = request.form.get("contactPerson")
+            releaseContactPerson = User.select().where(User.username == releaseContactUsername).get()
```
BrianRamsay: This can be shortened to `User.get(User.username == releaseContactUsername)`

app/templates/main/laborReleaseForm.html
```diff
@@ -70,6 +70,18 @@ <h1> Labor Release Form </h1>
...
       <!-- All of the fields on the right -->
       <div class= "floatright" >
+        <div class="form-group">
+            <label><span class="tooltip-right" data-tooltip="Required">Contacted Labor Office Member<span class="required-mark">*</span></span></label>
+            <div class='form-text '><small>Note: you are required to contact the labor office before submitting the form </small></div>
+          <div>
+            <select class="selectpicker" name="contactPerson" id="contactPerson" title="Name of the person you contacted from labor office" data-width="100%">
+              {% for admin in laborAdmins %}
+                {% set user = admin.supervisor if admin.supervisor else admin.student %}
+                <option value="{{admin.username}}">{{user.FIRST_NAME}} {{user.LAST_NAME}} </option>
```
BrianRamsay: User already has properties that can get a name from the supervisor or student record, so we should use those. In this case you could use `{{admin.fullName}}`

app/static/css/laborReleaseForm.css
```diff
@@ -44,7 +44,13 @@ textarea {
...
   box-sizing: border-box;         /* Opera/IE 8+ */
 }
-
+.space {
+  position: relative;
+  top: 1rem;
+}
+.studentspace {
+  margin-top: 21px;
+}
```
BrianRamsay: Instead of very specific style changes, I think it would be better to bump the left column down by including blank text above the first label. That way right and left will always match, if you use the same markup as the note on the right.

BrianRamsay (changes requested):

BrianRamsay:
If you aren't implementing the email feature, make sure that issue is created and at the top of the backlog so we see it.

app/controllers/main_routes/laborReleaseForm.py
BesherKitaz: Done.

app/templates/main/laborReleaseForm.html
BesherKitaz: Done. The reason why we had the above weird logic is because we did not see the fullName field in the database. Now I realized that it only exists in the models, not the actual database since it is defined as a function with a property decorator

app/static/css/laborReleaseForm.css
BesherKitaz: Removed and added a </br> tag within a div that has 'form-group' class. This makes it align perfectly. The div is necessary for the break to work. I don't why but it breaks when I removed it.

app/controllers/main_routes/laborReleaseForm.py
```diff
@@ -26,6 +26,7 @@ def laborReleaseForm(laborStatusKey):
             return redirect('/laborHistory/' + currentUser.student.ID)
 
     forms = LaborStatusForm.select().distinct().where(LaborStatusForm.laborStatusFormID == laborStatusKey)
+    laborAdmins = User.select().where(User.isLaborAdmin == 1)
```
BrianRamsay: Let's order the admins by first or last name

BrianRamsay (changes requested):
We should put the Contacted User into the display of the release form in the Labor History modal.

Screenshot 2025-07-22 at 3 57 50 PM

app/controllers/main_routes/laborReleaseForm.py
BesherKitaz: We added the contact person from the labor dep. into the model. See picture:

image

Also, now the list of labor Admins is ordered by First name (and then last name). Lastly, only labor admin how are professional staff are in the drop down list now.

BrianRamsay (approved):

BrianRamsay (approved):

BrianRamsay (approved):

PR #498, opened 2025-07-14
Pending Form Errors

GalinaP7:
**Isssue**
-Labor Dept. student workers get the pending forms error. Labor Dept. students should be able to use pending forms
-There is nothing in the database that can indicate or keep track of Labor Dept. Student Workers.

**Changes**
-Created the isLaborDepartmentStudent property that keeps track of whether the current user is a student who works in the labor department
-Used the isLaborDepartmentStudent property in allPendingForms and sidebar to apply that condition to the already existing ones. 
-Added a temporary department (labor dep), student, and user to demo data to test a user that fit the criteria (works in labor dep as a student).

**Testing**
-Run pytest -v tests/code/test_models.py to check that all the tests pass.
-Then go to the secret_config.ymal file and change the user on line 29 between someone who is a labor dept. student worker (test1), staff/admin (heggens), and just a student (jamalie)
-After switching user, reset the database from test and flask run, only test1 user should view Pending Forms (aside from those with admin/staff access).


app/controllers/admin_routes/allPendingForms.py
```diff
@@ -23,6 +23,7 @@
 from app.models.historyType import HistoryType
 from app.models.department import Department
 from app.controllers.main_routes.download import CSVMaker
+from app.models.user import User
```
BrianRamsay: Doesn't look like this import is needed anymore

app/models/user.py
```diff
@@ -17,6 +21,40 @@ class User(baseModel):
...
+        self._ldsCache = None
+    
+    @property
+    def isLaborDepartmentStudent(self):
+        if self._ldsCache is not None:
+            return self._ldsCache
+       
+        if not self.student:
+            self._ldsCache = False
+            return False
```
BrianRamsay: I would move this block above the first one in this function, and then you don't need the return because the next conditional will return the cache value. That way you have one fewer exit from the function, which makes it a little easier to work with.

app/models/user.py
```diff
@@ -17,6 +21,40 @@ class User(baseModel):
...
+        labor_status_forms = (
+            LaborStatusForm
+            .select()
+            .join(Department, on=(LaborStatusForm.department == Department.departmentID))
+            .where(
+                (LaborStatusForm.studentSupervisee == self.student) &
+                (LaborStatusForm.startDate <= today) &
+                (LaborStatusForm.endDate >= today) &
+                (Department.isActive == True) &
+                (fn.BINARY(Department.DEPT_NAME) == "Labor Department") # comparison case sensetive.
```
BrianRamsay: Why are we needing to use the BINARY function to work with this field?

app/models/user.py
```diff
@@ -17,6 +21,40 @@ class User(baseModel):
...
+            )
```
BrianRamsay, line 69: We need to make sure that students who are released from their position are excluded from this query. And also that this labor status form is approved. The first can be a subselect in the where clause, and the second a join with FormHistory (I think)

BrianRamsay (changes requested):

app/models/user.py
Arohasina: The BINARY forces the comparison to be case-sensitive by treating the string as a binary value, which is why we added it there. Or should we remove it?

app/templates/sidebar.html
```diff
@@ -62,7 +86,7 @@ <h4>Administration</h4>
                 <div id="admin" class='panel-collapse {% if request.path[:6] != "/admin"%} collapse {% endif %}'>
                   {% if currentUser.isLaborAdmin %}
                   <ul id="admindropdown" class="panel panel-default">
-                    <a href="/admin/pendingForms/pendingLabor"><li class="panel-heading {% if request.path == "/admin/pendingForms/pendingLabor" or request.path == "/admin/pendingForms/pendingAdjustment" or request.path == "/admin/pendingForms/pendingOverload" or request.path == "/admin/pendingForms/pendingRelease" %} active {% endif %}" tabindex="9"><h4>Pending Forms</h4></li></a>
+                     <a href="/admin/pendingForms/pendingLabor"><li class="panel-heading {% if request.path == "/admin/pendingForms/pendingLabor" or request.path == "/admin/pendingForms/pendingAdjustment" or request.path == "/admin/pendingForms/pendingOverload" or request.path == "/admin/pendingForms/pendingRelease" %} active {% endif %}" tabindex="9"><h4>Pending Forms</h4></li></a>
```
BesherKitaz, line 89: There seems to be a graying out of the "Pending Forms" when it is selected because of the "%} active {% "  in the css above. There should be a uniformity to the design. 
image

Arohasina: image
It is also greyed out on other tabs when selected. so I think it is consistent

app/models/user.py
```diff
@@ -17,6 +22,110 @@ class User(baseModel):
...
+            .join(Department, on=(LaborStatusForm.department == Department.departmentID))
+            .join(FormHistory, on=(FormHistory.formID == LaborStatusForm.laborStatusFormID))
+            .where(
+                (LaborStatusForm.studentSupervisee == self.student) &
+                (LaborStatusForm.startDate <= today) &
+                (LaborStatusForm.endDate >= today) &
+                (Department.isActive == True) &
+                (fn.BINARY(Department.DEPT_NAME) == "Labor Department") & 
+                (FormHistory.status == approved_status) &  
+                ~(LaborStatusForm.laborStatusFormID.in_(released_forms))  # Exclude released students
```
BrianRamsay: This will be easier for others to read if you use `.not_in()` instead of `~` and `in_`

app/models/user.py
```diff
@@ -17,6 +22,110 @@ class User(baseModel):
...
+        from app.models.status import Status
+        today = date.today()
+
+        # Get the approved status record
+        try:
+            approved_status = Status.get(Status.statusName == "Approved")
+        except Status.DoesNotExist:
+            # If no approved status exists, no forms can be approved
+            self._ldsCache = False
+            return self._ldsCache
```
BrianRamsay: You don't need to get a status object to use later. It's not the best database design, but lsf uses the status text as the keys, so you can do stuff like `FormHistory.status == "Approved"` in your query.

app/models/user.py
```diff
@@ -17,6 +22,110 @@ class User(baseModel):
...
+    #                     (LaborStatusForm.endDate >= today) &
+    #                     (Department.isActive == True) &
+    #                     (fn.BINARY(Department.DEPT_NAME) == "Labor Department") & # comparison case sensetive.
+    #                     (FormHistory.status == approved_status) & # Ensure form is approved
+    #                     ~(LaborStatusForm.laborStatusFormID.in_(released_forms))  # Exclude released students
+    #                 )
+    #             )
+            
+    #             self._ldsCache = labor_status_forms.exists()
+    #             return self._ldsCache
```
BrianRamsay: If we don't need the code anymore, we can remove it rather than comment it out.

app/models/user.py
```diff
@@ -17,6 +22,110 @@ class User(baseModel):
...
+                (FormHistory.releaseForm.is_null(False)) &  # Has a release form
+                (FormHistory.status == approved_status)     # Release form is approved
+            )
+        )
+
+        labor_status_forms = (
+            LaborStatusForm
+            .select()
+            .join(Department, on=(LaborStatusForm.department == Department.departmentID))
+            .join(FormHistory, on=(FormHistory.formID == LaborStatusForm.laborStatusFormID))
```
BrianRamsay: This join also needs to have a history type included (or it can be in the WHERE clause), because we will be getting other forms related to this one (like overload, adjustment, release, etc), not just the main contract.

BrianRamsay (changes requested):
Nice work! A couple of tweaks and we should be there.

app/models/user.py
```diff
@@ -17,6 +19,55 @@ class User(baseModel):
...
+            .join(FormHistory, on=(
+            (FormHistory.formID == LaborStatusForm.laborStatusFormID) &
+            (FormHistory.historyType == "Labor Status Form")  # Filter by Labor Status history type
+        ))
+            .where(
+                (LaborStatusForm.studentSupervisee == self.student) &
+                (LaborStatusForm.startDate <= today) &
+                (LaborStatusForm.endDate >= today) &
+                (Department.isActive == True) &
+                (fn.BINARY(Department.DEPT_NAME) == "Labor Department") &
```
BrianRamsay: There's no department named "Labor Department", and the name might change soon anyway. We should probably use ACCOUNT and ORG so a name change doesn't break the query. Let's use ACCOUNT 6740 and ORG either 4022 or 4187.

app/models/user.py
```diff
@@ -17,6 +19,55 @@ class User(baseModel):
...
+        from app.models.formHistory import FormHistory
+        today = date.today()
+
+        # Subquery to check if student has an approved release form
+        released_forms = (
+            FormHistory
+            .select(FormHistory.formID)
+            .where(
+                (FormHistory.releaseForm.is_null(False)) &  # Has a release form
+                (FormHistory.status == "Approved")     # Release form is approved
```
BrianRamsay: Let's go ahead and add in a join to `LaborStatusForm` and check `studentSupervisee` as well, so our list is release forms for the student, and not release forms for every student ever.

BrianRamsay (changes requested):
Sorry it took so long!

BrianRamsay (approved):

PR #500, opened 2025-07-15
fixed Student Search results to show preferred name not legal first name

Arohasina:
[Isssue #499](https://github.com/BCStudentSoftwareDevTeam/lsf/issues/499) 
-Student Search results should show the preferred name not legal first name

Changes
-when searching for a student, it now shows their preferred name not their legal name. 
image

Testing
-flask run
-Under the Student Search tab, look up a student who has a preferred name set up. For example: Jaylin James, preferred name is Jay.

### DIFF
--- a/app/templates/main/formHistory.html
+++ b/app/templates/main/formHistory.html
@@ -12,7 +12,7 @@
 <a class="skipContent" href="#downloadButton" tabindex="1">Click to Skip</a>
 <div align= "center">
   <h1>Labor History</h1>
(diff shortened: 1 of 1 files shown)


BesherKitaz:
Looks great! Everything is working out and the code looks clean

BrianRamsay:
This change actually introduces a bug to code that already works. `preferred_name` is not guaranteed to be set. `FIRST_NAME` is a property of the `Student` class that already uses `preferred_name` and then uses `legal_name` if `preferred_name` doesn't exist, so the code was working as intended originally.

Issue #501, opened 2025-07-15
Download options


Karina-Agliullova:
All Download options are broken

PR #509, opened 2025-07-21
Download button issue

RueHaile:
[Resolved issue](https://github.com/BCStudentSoftwareDevTeam/lsf/issues/501)

_Fixed_: All Download options are broken.

_Issue_: The download button was not working for the Labor History page. 

image

Test: 
1. Go to the LSF.
2. Log in as a student or log in as an admin, and then supervisor portal, followed by picking a student via the search(put the check buttons checked and keep the selection vague so that there are students)
3. Attempt to download the labor history.

### DIFF
--- a/app/controllers/main_routes/laborHistory.py
+++ b/app/controllers/main_routes/laborHistory.py
@@ -62,7 +62,7 @@ def laborhistory(id):
         
         authorizedForms = Term.order_by_term(list(authorizedForms.objects()), reverse=True)
 
-        laborStatusFormList = ','.join([str(form.formID.laborStatusFormID) for form in studentForms])
+        laborStatusFormList = ','.join([str(form.formID) for form in studentForms])
         return render_template('main/formHistory.html',
     				            title=('Labor History'),
                                 student = student,
@@ -82,13 +82,15 @@ def downloadFormHistory():
     This function downloads the created excel sheet of the history from the page.
     """
     try:
-        data = request.form
-        print("Data received for download: ", data)
-        historyList = data["listOfForms"].split(',')
-        print(historyList, "History List")
-        excel = CSVMaker("studentHistory", historyList, includeEvals = True)
-        return send_file(excel.relativePath, mimetype='text/csv', as_attachment=True, attachment_filename=excel.relativePath.split('/').pop())
-    except:
+        dataIds =  request.form.getlist('listOfFormIDs[]')
+        laborStatusForms = FormHistory.select().where(FormHistory.formID.in_(dataIds))
+
+        excel = CSVMaker("studentHistory", laborStatusForms, includeEvals=False)
+        download_url = url_for('static', filename='files/studentHistory.csv')
+        return jsonify({"download_url": download_url})
+
+    except Exception as e:
+        print("Error",e)
         return render_template('errors/500.html'), 500
 
 @main_bp.route('/laborHistory/modal/<statusKey>', methods=['GET'])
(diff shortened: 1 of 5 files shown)


app/static/js/laborhistory.js
```diff
@@ -36,6 +36,34 @@ $("#modal").on('transitionend', function(){ //Had to search for css element visi
...
+    url: '/laborHistory/download',
+    method: 'POST',
+    data: { "listOfFormIDs": historyList },
+    success: function(response) {
+      // Handle successful download
+      window.location.href = response.download_url;
+    },
+    error: function() {
+      // Handle download error
+      console.log("Download failed");
```
BrianRamsay, line 62: The user should see some indication that the download failed

app/controllers/main_routes/laborHistory.py
```diff
@@ -61,7 +61,7 @@ def laborhistory(id):
         
         authorizedForms = Term.order_by_term(list(authorizedForms.objects()), reverse=True)
 
-        laborStatusFormList = ','.join([str(form.formID.laborStatusFormID) for form in studentForms])
+        laborStatusFormList = ','.join([str(form.formID) for form in studentForms])
```
BrianRamsay, line 64: The fact that this works with no changes to the template probably means it's not being used. You can remove it, if so.

app/controllers/main_routes/laborHistory.py
```diff
@@ -81,11 +81,15 @@ def downloadFormHistory():
...
     try:
-        data = request.form
-        historyList = data["listOfForms"].split(',')
-        excel = CSVMaker("studentHistory", historyList, includeEvals = True)
-        return send_file(excel.relativePath, mimetype='text/csv', as_attachment=True, attachment_filename=excel.relativePath.split('/').pop())
-    except:
+        dataIds =  request.form.getlist('listOfFormIDs[]')
+        laborStatusForms = FormHistory.select().where(FormHistory.formID.in_(dataIds))
+
+        excel = CSVMaker("studentHistory", laborStatusForms, includeEvals=False)
```
BrianRamsay, line 87: We are referring to FormHistory objects and ids as 'forms' a few different places. That's confusing (and we do it a lot in LSF). Let's be more explicit about what is being stored and passed around.

app/static/js/laborhistory.js
```diff
@@ -36,6 +36,34 @@ $("#modal").on('transitionend', function(){ //Had to search for css element visi
...
+  e.preventDefault();
+  let table = $('#positionTable');
+  let historyList = [];
+
+  table.find('tr').each(function(){
+    let formID = $(this).find('td').attr('id');
+    if(formID){
+      historyList.push(formID);
+    }
+  });
```
BrianRamsay, line 49: Consider using `.map()` instead of `.each()` to build `historyList` directly from the list of rows. It might be more self-explanatory.

BrianRamsay (changes requested):

app/controllers/main_routes/laborHistory.py
BesherKitaz: Yo that's dope! 
I mean, yes, you are right. Removed it now.

app/static/js/laborhistory.js
BesherKitaz: We changed it to use map. We also used filter to replace the if statement.

app/controllers/main_routes/laborHistory.py
BesherKitaz: We changed dataIds to formIds. We also changed laborStatusForm to laborStatusFormObjs to make it more explicit that they are objects. Let me know if that is sufficient for clarity.

app/static/js/laborhistory.js
BesherKitaz: done.

BrianRamsay:
This had been fixed directly in production.

PR #529, opened 2025-09-08
Dynamic search and load time optimization

### LINKED ISSUE #502: Supervisor Portal
Supervisor Portal takes too long to load. There are unnecessary queries running

JohnCox2211:
Closes #502 

First, we cut down unnecessary double-querying in the Supervisor Portal and removed a few hanging + unused queries.

Then, we added a dynamic search function for the supervisor portal for all of the dropdown boxes. This is so that when the page is loaded, the boxes with a lot of options are not prefilled with a large amount of HTML that's largely unused, as supervisors scarcely scan through the longer dropdowns, especially student names.

However, the Departments dropdown's functionality remains the same. Due to the small number of items in the dropdown, it's faster for users to simply scan through with their eyes and click, instead of searching.

### Testing 
- You may test the dynamic search on any of the boxes in supervisor portal. The only box that will be automatically populated with anything but the default 'all' options will be departments.
- Compare load times and number of queries generated

### DIFF
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -16,8 +16,6 @@ from app.logic.search import getDepartmentsForSupervisor
 from app.login_manager import require_login, logout
 from app.logic.getTableData import getDatatableData
 from app.logic.banner import Banner
-from app.models.user import User # remove later?
-from app.logic.tracy import Tracy # remove later?
 
 @main_bp.route('/logout', methods=['GET'])
 def triggerLogout():
@@ -90,6 +88,7 @@ def SupervisorPortalSearch():
     """
     Returns a list of users that match a given string
     """
+    print('liveSearch called')
     def searchSupervisorPortal(searchType, userInput):
         currentUser = require_login()
         if currentUser.isLaborAdmin or currentUser.isFinancialAidAdmin or currentUser.isSaasAdmin:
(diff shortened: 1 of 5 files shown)


RueHaile:
It is a good implementation of the dynamic search. It has done the task of cutting down the loading time (optimization).  Good job.

app/controllers/main_routes/main_routes.py
```diff
@@ -112,7 +84,59 @@ def downloadSupervisorPortalResults():
     )
     return send_file(excel.relativePath, as_attachment=True, attachment_filename=excel.relativePath.split('/').pop())
 
-@main_bp.route('/lsf/<formHistoryId>/submitToBanner', methods=['GET'])
+@main_bp.route('/supervisorPortal/liveSearch', methods=['GET'])
+def SupervisorPortalSearch():
+    """
+    Returns a list of users that match a given string
+    """
+    def searchSupervisorPortal(searchType, userInput):
```
BrianRamsay: I'm not a fan of the method being defined inside another method approach here. This should go in logic/search.py

BrianRamsay (dismissed):
The form doesn't remember the option chosen in the new search dropdowns after a page reload. It looks like the data returned is correctly reflecting it, but the supervisor (for example) doesn't show up as selected. If you check another box and Submit you get a datatables error.

BrianRamsay (approved):

PR #539, opened 2025-09-16
added flash message

### LINKED ISSUE #521: Flash Message after form approval
There is no flash message after the labor status form has been approved 

Image

bakobagassas:
fixes #521 

Flash message after form approval

**Changes** 

- Added a flash message if the final approval is a success

**Testing** 

- There has been a problem related to sending the emails when we want to approve a form. We cannot see the message unless the email is sent. Otherwise everything should be fine with the flash message.


BrianRamsay (changes requested):
Maybe I'm missing something, but it looks like the page reloads (from the pending forms page, anyway), and so the javascript flash disappears before you can see it. We probably need a flash from the controller so it shows on the next page load.

PR #540, opened 2025-09-16
Administrator remove fix

### LINKED ISSUE #530: Administrators Being Removed When They Shouldn't Be
### Background
- Go to the Manage Administrators page under the Administration tab on the sidebar
- When adding a user from the select user box, you are able to type a name, if you type a name less than 3 characters, you are able to hit 'Add' to submit it 

### Issue
- If you add someone in the Current Labor Admins box next to the remove button and leave the name there, if you try to submit a name less than 3 characters in the box above and then hit the 'Add' button, the person in the bottom box will be removed as an admin. 

### Testing

**The Page**
Image

**Add someone to the bottom box**
Image

**Type a < 3 character name in the top box and press enter** 
Image

**click 'Add',  then click 'Select' in the modal**
Image

**The person in the bottom box has been removed**

JohnCox2211:
Fixes issue #530 

Refactored code that deals with adding and removing admins in order to make it more robust for possible future modifications and prevents improper admin removals that are currently possible.


bakobagassas:
On your branch, when I add a name that is <3 letters and click enter and remove, instead of removing them, I receive a flash message saying they have been added as a labor admin. This is weird because I tried the same thing I did another time and it did not happened then I did it a third time and the problem happened that time.

bakobagassas:
Also when I click to add alex for example: (1) I put 3 characters, the name appeared (2) I removed one character, the name was still there (3) but I am still unable to actually click on his name although it is there on my screen.

JohnCox2211:
> Also when I click to add alex for example: (1) I put 3 characters, the name appeared (2) I removed one character, the name was still there (3) but I am still unable to actually click on his name although it is there on my screen.

Upon testing, this bug that  you mentioned also appears in the organizingtest branch, so it is not new to our branch. We will, however, incorporate this change into our branch

JohnCox2211:
> On your branch, when I add a name that is <3 letters and click enter and remove, instead of removing them, I receive a flash message saying they have been added as a labor admin. This is weird because I tried the same thing I did another time and it did not happened then I did it a third time and the problem happened that time.

I was unable to replicate this bug, could you show me next time we are both in the programmer space?

BrianRamsay (changes requested):
`getUser` is being imported from a file in `logic/`, but you've added it back in to the controller also. Was that a merge conflict issue?

JohnCox2211:
> `getUser` is being imported from a file in `logic/`, but you've added it back in to the controller also. Was that a merge conflict issue?

Resolved this issue, deleted duplicate function in controller

app/controllers/admin_routes/adminManagement.py
```diff
@@ -47,35 +47,39 @@ def adminSearch():
...
-        adminFlashMessage(oldAdmin, 'removed', 'Financial Aid')
+    actionMap = {
+    "addAdmin":                {"type": "labor", "action": "add", "pretty": "Labor"},
+    "removeAdmin":             {"type": "labor", "action": "remove", "pretty": "Labor"},
+    "addFinancialAidAdmin":    {"type": "finAid", "action": "add", "pretty": "Financial Aid"},
+    "removeFinancialAidAdmin": {"type": "finAid", "action": "remove", "pretty": "Financial Aid"},
+    "addSAASAdmin":            {"type": "saas", "action": "add", "pretty": "SAAS"},
+    "removeSAASAdmin":         {"type": "saas", "action": "remove", "pretty": "SAAS"},
+    }
+    for key, meta in actionMap.items():
```
BrianRamsay: You don't need to loop through the dictionary if you only have one key. You can just select the values from the dictionary directly:
`meta = actionMap.get(request.form.get(key))`

app/controllers/admin_routes/adminManagement.py
```diff
@@ -47,35 +47,39 @@ def adminSearch():
...
+        if request.form.get(key):
+            user = getUser(key)
+            
+            # pick addAdmin or removeAdmin dynamically
+            (addAdmin if meta["action"] == "add" else removeAdmin)(user, meta["type"])
```
BrianRamsay: Let's keep this as a multi-line if/else

app/static/js/adminManagement.js
```diff
@@ -43,56 +43,46 @@ function liveSearch(selectPickerID, e) {
...
+
+  const cfg = config[button];
+
+  const select = $(cfg.select);
+  if (!select.val()) {
+    // no user selected
+    $("#flash_container").html(
+      '<div class="alert alert-danger" role="alert" id="flasher">Please select a user.</div>'
+    );
+    $("#flasher").delay(3000).fadeOut();
```
BrianRamsay: Let's use our javascript flash message function

app/static/js/adminManagement.js
```diff
@@ -43,56 +43,46 @@ function liveSearch(selectPickerID, e) {
...
+  const config = {
+    add:   { select: "#addlaborAdmin",       form: "#laborAdminForm",   header: "Labor Administrators",       action: "add",    noun: "Labor Administrator" },
+    add1:  { select: "#addFinAidAdmin",      form: "#finAidAdminForm",  header: "Financial Aid Administrators", action: "add",    noun: "Financial Aid Administrator" },
+    add2:  { select: "#addSaasAdmin",        form: "#SAASAdminForm",    header: "SAAS Administrators",        action: "add",    noun: "SAAS Administrator" },
+    remove:{ select: "#removelaborAdmin",    form: "#laborAdminForm",   header: "Labor Administrators",       action: "remove", noun: "Labor Administrator" },
+    remove1:{ select: "#removeFinAidAdmin",  form: "#finAidAdminForm",  header: "Financial Aid Administrators", action: "remove", noun: "Financial Aid Administrator" },
+    remove2:{ select: "#removeSaasAdmin",    form: "#SAASAdminForm",    header: "SAAS Administrators",        action: "remove", noun: "SAAS Administrator" }
+  };
+
+  const cfg = config[button];
```
BrianRamsay: `config` and `cfg` are too close to each other. let's come up with more descriptive names, particularly for `cfg`.

BrianRamsay (changes requested):

app/static/js/adminManagement.js
JohnCox2211: Renamed config to formButtons and cfg to adminForm. These names seem to be better descriptors.

app/controllers/admin_routes/adminManagement.py
JohnCox2211: Changed how I was handling the logic there and no longer need that loop there.

BrianRamsay:
I added a commit to change the number of queries on page load from 805 to 10.

BrianRamsay (approved):

Issue #504, opened 2025-07-15
Expiration script


Karina-Agliullova:
Create expiration script for form approval

BrianRamsay:
![](https://github.trello.services/images/mini-trello-icon.png) [Create expiration script for form approval.](https://trello.com/c/SwIK7qQy/900-create-expiration-script-for-form-approval)

PR #541, opened 2025-09-23
Form expiration script

RueHaile:
[Pull request on git hub ](https://github.com/BCStudentSoftwareDevTeam/lsf/issues/504)
Issue number: #541

**Issue** 
Create an expiration script for form approval. 
When a form was expiring, it was not communicated to the student or the supervisor. 

**Fixed:** 
Both the student and the supervisor are being notified when a labor status form is expiring. 
Added a script that checks if a form expires and sends an email. 

**Test:**
Go to the secret config and paste the statement below the financial aid email as follows (don't forget to set your password and your email in the code below):
1. 
ALWAYS_SEND_MAIL: True
MAIL_SERVER: "smtp.gmail.com" #use smtp.gmail.com for gmail or smtp.office365.com for Berea
MAIL_PORT: 465 #: use port 465 for gmail use 587 for Berea
MAIL_USERNAME: '**put your personal email here** ' #the email of the sender
MAIL_PASSWORD: '**Put the password that you have generated in here**' #Add the password of the sender 
**For the password above, go to your Gmail and turn on your 2-step verification, followed by setting an  app passowrd, you will put that password in the statement above**
MAIL_USE_TLS: False # False for gmail, True for berea
MAIL_USE_SSL: True # True for gmail, False for berea
MAIL_DEFAULT_SENDER: '**put your personal email here** @' 
MAIL_OVERRIDE_ALL:  '**put your personal email here** @gmail.com' 
REPLY_TO_ADDRESS: "laborstatusform-aaaaazdxy4j2lsinl67ikumy4y@studentprogrammers.slack.com"

2. Reset data database by: ./database/reset_database.sh from-backup
3. Run the script with python3 ./scripts/automatedFormExpiration.py

**_success == When the script runs successfully, you should get an email._**


scripts/automatedFormExpiration.py
```diff
@@ -0,0 +1,88 @@
...
+                continue
+
+            lsfHistory = latest_history.formHistoryID
+            emailer = emailHandler(lsfHistory)
+            emailer.laborStatusFromExpired()
+
+        except Exception as e:
+            import traceback
+            print(f"Error while processing LSF#{form.laborStatusFormID}: {e}")
+            traceback.print_exc()
```
ojmakinde: Why are we using traceback here? It's also incredibly unconventional to import a package under an exception. I assume this was used during debugging. Additionally, we don't need to catch an exception if all it's gonna do is simply print out an error. The original error should simply be left in place without wrapping in a try/except block

app/logic/emailHandler.py
```diff
@@ -308,11 +344,13 @@ def sendEmail(self, template, sendTo):
         formTemplate = template.body
         formTemplate = self.replaceText(formTemplate)
         if sendTo == "student":
+            print("here", template.subject, self.supervisorEmail)
             message = Message(template.subject,
                 recipients=[self.studentEmail])
             recipient = 'Student'
         elif sendTo == "secondary":
             if self.term.isBreak:
+                print("here", template.subject, self.supervisorEmail)
```
ojmakinde: the surrounding conditionals are littered with print statements, please remove them when you can

ojmakinde (changes requested):

app/logic/emailHandler.py
```diff
@@ -16,6 +16,8 @@
 from app import app
 import os
 from datetime import datetime, date
+import traceback
```
ojmakinde: please remove this import, since it's not being used anywhere

app/logic/emailHandler.py
```diff
@@ -136,8 +138,40 @@ def laborStatusFormSubmitted(self):
         else:
             self.checkRecipient("Labor Status Form Submitted For Student",
                           "Primary Position Labor Status Form Submitted")
-
-            
+    # This method was commented out because it was causing issues with the automated script that checks for expired formsdef laborStatusFromExpired(self):
```
ojmakinde: what method was commented out?

app/logic/emailHandler.py
```diff
@@ -136,8 +138,40 @@ def laborStatusFormSubmitted(self):
...
+        expired = self.laborStatusForm.isExpired
+        if not expired:
+            return
+        supervisorTemplate = EmailTemplate.get_or_none(
+            EmailTemplate.purpose == "Email when Labor Status Form is expired to Supervisor"
+        )
+        if not supervisorTemplate:
+            return
+        try:
+            today_val = date.today()  # works if EmailTracker.date is a DATE; if it's a string, cast to str below
```
ojmakinde: nit: can we use a better variable name?

app/logic/emailHandler.py
```diff
@@ -136,8 +138,40 @@ def laborStatusFormSubmitted(self):
...
+            already_sent = (EmailTracker
+                            .select()
+                            .where(
+                                (EmailTracker.formID == lsfID) &
+                                (EmailTracker.subject == supervisorTemplate.subject) &
+                                (EmailTracker.date == today_val)  # if EmailTracker.date is a VARCHAR, use str(today_val)
```
ojmakinde: but you're not checking if emailtracker.date is a varchar and casting `today_val`, right? if it's a varchar, the current implementation is gonna raise an exception and send it regardless. why don't you first retrieve the emailtracker object for this form ID and compare against the date type? (varchar or datetime object)

app/logic/emailHandler.py
```diff
@@ -308,11 +342,13 @@ def sendEmail(self, template, sendTo):
         formTemplate = template.body
         formTemplate = self.replaceText(formTemplate)
         if sendTo == "student":
+            print("here", template.subject, self.supervisorEmail)
```
ojmakinde: it still seems like there's unnecessary print statements around the conditional. are these important?

scripts/automatedFormExpiration.py
```diff
@@ -0,0 +1,78 @@
...
+from datetime import date
+from app import app as flask_app
+from app.models.laborStatusForm import LaborStatusForm
+from app.models.formHistory import FormHistory
+from app.logic.emailHandler import emailHandler
+from app.models.emailTemplate import*
+from flask import has_request_context
+
+#emailhistory requires an active flask app context so we create one to supllement it
+BASE_URL = os.getenv("EXTERNAL_BASE_URL", "http://localhost:5000/")
```
ojmakinde: curious, what if my flask app is running on 8080 instead of 5000. will this work?

scripts/automatedFormExpiration.py
```diff
@@ -0,0 +1,78 @@
...
+        "formType":"Labor Status Form", 
+        "action":"Expired", 
+        "subject":'Labor Status Form Expired', 
+        "body":'<p>Dear <strong>@@Student@@</strong>,</p>   <p>This email is to notify that the Labor Status Form submitted by <strong>@@Creator@@</strong> for you has expired.</p>   <p>&nbsp;</p>   <p>Please notify your supervisor <strong>@@Student@@</strong> to reubmit the form.</p>   <p>Sincerely,<br>Labor Program Office<br>labor_program@berea.edu<br>859-985-3611</p>', 
+        "audience":"Student",
+    }
+}
+         
+def checkForTemplates():
+    for template in emailTemplates.values():
```
ojmakinde, line 39: if emailTemplates has only one item in it, why are we iterating over it?

scripts/automatedFormExpiration.py
```diff
@@ -0,0 +1,78 @@
...
+        if EmailTemplate.select().where(EmailTemplate.purpose == template["purpose"]).count() == 0:
```
ojmakinde: you don't have to check before inserting, i think. using an [upsert](https://docs.peewee-orm.com/en/latest/peewee/querying.html#upsert) should work here. right? @BrianRamsay 

additionally, you can simply unpack the dictionary here using a double asterisk. ([example](https://www.geeksforgeeks.org/python/packing-and-unpacking-arguments-in-python/))

scripts/automatedFormExpiration.py
```diff
@@ -0,0 +1,78 @@
...
+                .where(FormHistory.formID == form)
+                .order_by(FormHistory.formHistoryID.desc())
+                .first()
+            )
+            if not latest_history:
+                continue
+
+            lsfHistory = latest_history.formHistoryID
+            emailer = emailHandler(lsfHistory)
+            emailer.laborStatusFromExpired()
```
ojmakinde: nit: misspelling

app/logic/emailHandler.py
```diff
@@ -136,8 +138,40 @@ def laborStatusFormSubmitted(self):
         else:
             self.checkRecipient("Labor Status Form Submitted For Student",
                           "Primary Position Labor Status Form Submitted")
-
-            
+    # This method was commented out because it was causing issues with the automated script that checks for expired formsdef laborStatusFromExpired(self):
+    def laborStatusFromExpired(self):
```
ojmakinde: you misspelled Form. also, i think we should update the name to what it actually does. since this sends an email (according to the docstring), it should be renamed to something along those lines

ojmakinde (changes requested):

scripts/automatedFormExpiration.py
RueHaile: Yes, it is just one right now but it is a loop to make the code more flexable for the future.

scripts/automatedFormExpiration.py
RueHaile: good one.

BrianRamsay:
There are comments related to this PR in https://github.com/BCStudentSoftwareDevTeam/lsf/pull/546

MImran2002:
> There are comments related to this PR in #546
I have commented it these two issues are quite similar do you think we should have them separate or together

app/models/formHistory.py
```diff
@@ -10,7 +10,7 @@
 
 class FormHistory(baseModel):
     formHistoryID       = PrimaryKeyField()
-    formID              = ForeignKeyField(LaborStatusForm, on_delete="cascade")               # foreign key to lsf
+    formID              = ForeignKeyField(LaborStatusForm, on_delete="cascade", backref="lsf")               # foreign key to lsf
```
ojmakinde, line 13: https://github.com/BCStudentSoftwareDevTeam/lsf/pull/546#discussion_r2474559070

RueHaile:
> > There are comments related to this PR in #546
> > I have commented it these two issues are quite similar. Do you think we should have them separate or together

I second this because #546 has all the changes in here and more, as it is based on this pr.

ojmakinde (changes requested):

JohnCox2211:
When I ran the script that you provided, I got two duplicate emails. Both of the emails addressed the supervisor rather than one addressing the supervisor and the student. It is noteworthy that both of the emails also say that they are originally intended for the supervisor. Here are the two emails:

image


and 


image

MImran2002:
This PR is merged into Labor Release Form Admin

PR #544, opened 2025-10-13
Flash message after form approval

JohnCox2211:
Add a flash message to appear after selected pending forms have been approved. 

### TESTING

- Load up the all pending forms page
- Check one or more of the boxes next to the forms
- Go to the bottom of the page and hit the "Approve Selected Forms" button
- In the modal that appears, press the "Approve Selected" button 
- After the information is saved, a flash message should appear saying that the forms have been approved

Image for context:
image

### DIFF
--- a/app/static/js/allPendingForms.js
+++ b/app/static/js/allPendingForms.js
@@ -141,6 +141,7 @@ function finalApproval() { //this method changes the status of the lsf from pend
     contentType: 'application/json',
     success: function(response) {
       if (response && response.success) {
+          msgFlash("The selected forms have been approved.","success");
           $(".btn").prop("disabled", false);
           $(".close").prop("disabled", false);
           $("#approveModalButton").text("Approve");
--- a/app/static/js/allPendingForms.js
+++ b/app/static/js/allPendingForms.js
@@ -121,6 +121,10 @@ $('#approvalModal').on('hidden.bs.modal', function () {// Makes the close functi
 function approvalModalClose(){// on close of approval modal we are clearing the table to prevent duplicate data.
   $('#classTableBody').empty();
   labor_details_ids = [] // emptying the list, becuase otherwise will cause duplicate data.
+  if (window.approvalSuccess) {
+    msgFlash("The selected forms have been approved.","success");
+    window.approvalSuccess = false;
+  }
 }
 
 
@@ -141,7 +145,7 @@ function finalApproval() { //this method changes the status of the lsf from pend
     contentType: 'application/json',
     success: function(response) {
       if (response && response.success) {
-          msgFlash("The selected forms have been approved.","success");
+          window.approvalSuccess = true;
           $(".btn").prop("disabled", false);
           $(".close").prop("disabled", false);
           $("#approveModalButton").text("Approve");
--- a/app/static/js/allPendingForms.js
+++ b/app/static/js/allPendingForms.js
@@ -121,10 +121,6 @@ $('#approvalModal').on('hidden.bs.modal', function () {// Makes the close functi
 function approvalModalClose(){// on close of approval modal we are clearing the table to prevent duplicate data.
   $('#classTableBody').empty();
   labor_details_ids = [] // emptying the list, becuase otherwise will cause duplicate data.
-  if (window.approvalSuccess) {
-    msgFlash("The selected forms have been approved.","success");
-    window.approvalSuccess = false;
-  }
 }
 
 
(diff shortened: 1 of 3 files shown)


MImran2002:
When the modal appears and the approve selected is clicked it takes a couple of second but when the flask message appears, it is behind the modal and disappear first before the modal does. I think it is to do with the msgFlash function and it is in the base.js. I think you can tweak this line:    $("#flasher").delay(5000).fadeOut(); to make it last longer so that before the modal disappear the flash message will still be there or you can make it appear after the modal disappear.

app/static/js/allPendingForms.js
```diff
@@ -121,6 +121,10 @@ $('#approvalModal').on('hidden.bs.modal', function () {// Makes the close functi
 function approvalModalClose(){// on close of approval modal we are clearing the table to prevent duplicate data.
   $('#classTableBody').empty();
   labor_details_ids = [] // emptying the list, becuase otherwise will cause duplicate data.
+  if (window.approvalSuccess) {
+    msgFlash("The selected forms have been approved.","success");
```
MImran2002: I looked into it and realized even though msgflash appears after the function is called later on it is still underneath the modal because of z.index so I put a new parameter as a test for the msgFlash that will differentiate where it needs to show up after the modal or not so the msgFlash could look like this :     msgFlash("The selected forms have been approved.","success", true);

and because I want the flash message to show up I rely on the parameter yes or no if its yes we will append the flash message to the body so that when we raised the z.index it will not fall into the css stack but will appear over it. 
function msgFlash(flash_message, status, higher_priority=false) {
    var category = (status === "success") ? "success" : ((status === "fail") ? "danger" : "warning");
    if (higher_priority) {
        $("body").prepend('<div class="alert alert-'+ category +'" role="alert" id="flasher" style="position: fixed; top: 1rem; right: 1rem; z-index: 2000;">'+flash_message+'</div>');
        $("#flasher").delay(10000).fadeOut();
    } else {
    $("#flash_container").prepend('<div class="alert alert-'+ category +'" role="alert" id="flasher">'+flash_message+'</div>');
    $("#flasher").delay(5000).fadeOut();
    }
}

BrianRamsay (changes requested):
Does this work for you? It doesn't seem like we could ever get the flash message. The `runformSearchQuery` call on line 158 doesn't make sense, and fails, which means the page reloads every time, so the js flash message never shows.

We either need to replace `runformSearchQuery` with a function that actually reloads the table, or just get rid of the try/catch altogether and reload the whole page, which means we need a flash from the backend, not the frontend.

JohnCox2211:
> Does this work for you? It doesn't seem like we could ever get the flash message. The `runformSearchQuery` call on line 158 doesn't make sense, and fails, which means the page reloads every time, so the js flash message never shows.
> 
> We either need to replace `runformSearchQuery` with a function that actually reloads the table, or just get rid of the try/catch altogether and reload the whole page, which means we need a flash from the backend, not the frontend.

I've never got to test it so I would figure it doesn't work for me either. I had an initial solution that I got to test one time and it seemed to work but didn't when tested again, so the most recent change you saw isn't a finished or tested thing.

JohnCox2211:
> Does this work for you? It doesn't seem like we could ever get the flash message. The `runformSearchQuery` call on line 158 doesn't make sense, and fails, which means the page reloads every time, so the js flash message never shows.
> 
> We either need to replace `runformSearchQuery` with a function that actually reloads the table, or just get rid of the try/catch altogether and reload the whole page, which means we need a flash from the backend, not the frontend.

Added a flash message to the backend at the bottom of the updateStatus route

MImran2002:
image
Seems working now! I will merge it now

BrianRamsay (changes requested):
This works, but only because that try/catch is still failing. If we want to reload the page (which is fine), get rid of the failing try/catch and just location.reload() every time.

JohnCox2211:
> This works, but only because that try/catch is still failing. If we want to reload the page (which is fine), get rid of the failing try/catch and just location.reload() every time.

Commited this recommended change.

RueHaile:
> This works, but only because that try/catch is still failing. If we want to reload the page (which is fine), get rid of the failing try/catch and just location.reload() every time.

I agree, but merely having the ( location.reload() ) allows us to reload the page and perform the changes (in this case, allow the selected pending status form ), but the flash message won't disappear.  Unless we reload the page, it is common practice to have it fade out in time. 
I propose that we add the following to base.html

   <script>
        window.setTimeout(function() {
            $("#flash_container .alert").fadeTo(500, 0).slideUp(500, function() {
            $(this).remove();
            });
        }, 4000);
        </script>

  in the base.js so that it fades away as well.

BrianRamsay (approved):

PR #558, opened 2026-01-20
Labor History datetime Quick Fix

JohnCox2211:
datetime syntax was causing labor form modals to not work properly on Labor History pages because the use of .date for this function was deprecated. I also removed an unused line.


JohnCox2211:
[Labor History datetime Quick Fix # 558](https://trello.com/c/nCZXBtDG/1071-labor-history-datetime-quick-fix-558)

MImran2002 (approved):
student variable deleted is a variable not in used and datetime has been fixed

PR #563, opened 2026-01-29
Overload Information in Labor History Enhancement

### LINKED ISSUE #536: Admins should see Overload Information in Labor History
Currently when viewing a form in Labor History that went through the overload process, there is no visibility into the notes or approval process. Admins should be able to see the student reason and all notes for a form from within the Labor History view.

JohnCox2211:
Fixes issue #536 

### Changes:
Adds visibility into the overload form process for individual forms on the student Labor History page. This includes sections for notes on an overload rejection, the labor office contact for a form, the labor office's review date for the form, and the student's reason for the overload. 

### Testing:
**Student Search:**
    - Find a student with overload forms (I found Nicholas Hamilton to be a good example)
    - Open a form on the student's page with the overload status
    - If the form was approved or pending, there should be a section with a labor office contact, labor review date, and the student's reason for overload. 
    - If the form was denied, the above section should be included, as well as one with a rejection reason from the reviewer.

**Example** (Denied form with no contact listed):

image

**Example** (approved form with a contact listed):

image


app/templates/snips/studentHistoryModal.html
```diff
@@ -81,6 +81,62 @@
...
+          <div class="col-sm-3">{{form.reviewedDate.strftime('%m-%d-%Y')}}</div>
+          <div class="col-sm-5 ">{{form.historyType}}</div>
+          <div class="col-sm-4" align="left">{{form.status}}</div>
+        </div>
+        {% if "Denied" in form.status.statusName %}
+        <div class="leftMargin">
+          <div class="row">
+            <p class="col-sm-4 pTag">Reject Reason:</p>
+          <div class="col-sm-7">
+            <textarea class="overloadReasonTextArea" rows="2" cols="45" readonly>{{form.studentOverloadReason}}</textarea>
```
RueHaile: The presence of a textarea is misleading because it creates an illusion of interactivity. A simple div with customizations for scrollability and space flexibility would be optimal.

JohnCox2211:
### Known issues as of 2/5/26:

- WHEN MAKING AN OVERLOAD BY ADJUSTING A FORM
        adjustment pending form for an overload shows "{ "Success": false }"

- adjustment pending form shows "{ "Success": false }" after only changing the notes
                ***turns out this is also an issue on development

- pending adjustment forms tab in pending forms gives this error:
	"An unexpected error has occurred
	The administrator has been notified. Sorry for the inconvenience!"
                ***turns out this is also an issue on development

- WHEN MAKING AN OVERLOAD FORM FROM New LSF PAGE
        Overload Pending form shows "{ "Success": false }"

Kafui123:
I think it would be a good idea to bold these texts: the labor office contact, the labor review date, and the reason for overload. 

image

app/controllers/main_routes/laborHistory.py
```diff
@@ -131,8 +131,13 @@ def populateModal(statusKey):
         currentUser = require_login()
         if not currentUser:                    # Not logged in
             return render_template('errors/403.html'), 403
-        forms = (FormHistory.select().join(LaborReleaseForm, join_type=JOIN.LEFT_OUTER)
-                            .where(FormHistory.formID == statusKey).order_by(FormHistory.createdDate.desc(), FormHistory.formHistoryID.desc()))
+        forms = (FormHistory.select()
+                            .join(LaborReleaseForm, join_type=JOIN.LEFT_OUTER)
+                            .switch(FormHistory).join(OverloadForm, JOIN.LEFT_OUTER, on=(FormHistory.overloadForm == OverloadForm.overloadFormID))
+                            .switch(OverloadForm).join(User, JOIN.LEFT_OUTER, on=(OverloadForm.laborApprover == User.userID))
```
Kafui123: I think that the (.switch(overloadForm ... line 137) and the (.switch(User)... line 138) could be redundant. Since you already have a .join(overloadForm) and a .join(User), Peewee automatically switches the current model to overloadForm and User.

app/templates/snips/studentHistoryModal.html
```diff
@@ -211,6 +215,32 @@
         </div>
       {% endif %}
 
+        <div class="smallDataMargin">
+        <div class="row" style="padding-top:5px">
+          <p class="col-sm-4 releaseTopRows pTag">Labor Office Contact:</p>
+          <div class="col-sm-7" align="left">
+            <p>
+              {% if form.overloadForm.laborApprover_id and (form.overloadForm.laborApprover.supervisor.FIRST_NAME or form.overloadForm.laborApprover.supervisor.LAST_NAME) %}
```
Kafui123: it would be cleaner and more concise if you stored the supervisor in a variable and then checked if the supervisor exists, as well as the first and last name .... i.e 

{% set supervisor = form.overloadForm.laborApprover.supervisor %}

{% if supervisor and (supervisor.FIRST_NAME or supervisor.LAST_NAME %}

JohnCox2211:
> I think it would be a good idea to bold these texts: the labor office contact, the labor review date, and the reason for overload.
> 
> image

The reason why I ended up not bolding the text was to match the appearance of the additional information already in place for release forms. I do agree that both would look better if they stood out more, so I have implemented this change for both as well the extra info popup for adjusted forms.

JohnCox2211:
**Possible conflicting semi-related issues you may run into:**
- When adjusting a form, the department ends up changing to some default that is not the department connected to the form, so it will appear as None. I have added error handling to catch this problem so that the labor history modal still appears to allow testing of this issue. 
- When adjusting an existing labor status form for it to become an overload form, the student is unable to accept the form in the pre-student-approval phase. To avoid running into this in testing, simply create a labor status form as an overload form. 

Note: both of the above issues are in the development branch as well and should be fixed in another issue.

RueHaile:
image
 The box text area is missleading it seems interactive when in fact it is not. We can have a better UI where it is merely a display, and the presentation is not misleading. Is there a reason that is the case?

JohnCox2211:
> image The box text area is missleading it seems interactive when in fact it is not. We can have a better UI where it is merely a display, and the presentation is not misleading. Is there a reason that is the case?

The reason this textbox is kept this way is that it is the same as the release form (which you screenshotted). I kept that format for the overload forms for consistency. Though I agree that it would look better if it were disabled, so I implemented that change for the release and overload text areas

app/controllers/main_routes/laborHistory.py
```diff
@@ -166,10 +171,11 @@ def populateModal(statusKey):
...
 
                 if form.adjustedForm.fieldAdjusted == "department":
-                    newDepartment = Department.get(Department.ORG == newValue)
-                    oldDepartment = Department.get(Department.ORG == oldValue)
-                    form.adjustedForm.newValue = newDepartment.DEPT_NAME
-                    form.adjustedForm.oldValue = oldDepartment.DEPT_NAME
+                    newDepartment = Department.get_or_none(Department.ORG == newValue)
+                    oldDepartment = Department.get_or_none(Department.ORG == oldValue)
+                    form.adjustedForm.newValue = newDepartment.DEPT_NAME if newDepartment else "Unknown " + newValue
+                    form.adjustedForm.oldValue = oldDepartment.DEPT_NAME if oldDepartment else "Unknown " + oldValue
```
Kafui123, line 174: Could we possibly get rid of the "unknown" and just put the new value to make it cleaner?

JohnCox2211: This line is more of an error catching for another problem, and is meant just to allow the modal to open up when you encounter one of the known errors mentioned in a previous comment I made. This is what that ends up looking like:

image

JohnCox2211: These lines here are for error handling for other issues the could be encountered when testing for this PR and the wording follows other error handling in the same file. I feel like the additional label will be helpful for fixing the overall problem

app/templates/snips/studentHistoryModal.html
```diff
@@ -67,9 +67,9 @@
         {% if "Denied" in form.status.statusName %}
         <div class="leftMargin">
            <div class="row">
-             <p class="control-label col-sm-5 pTag" align="right">Reject Reason: </p>
+             <p class="control-label col-sm-5 pTag extraInfoRows" align="right">Reject Reason: </p>
```
MImran2002: I don't think we need to have a class for font-weight, I know lsf has very old bootstrap so using bootstrap is out of the picture. However, I found out the codebase uses <strong> so we can use that as I believe this class isn't added conditionally either so we can definitely use <strong>

app/static/css/studentHistoryModal.css
```diff
@@ -72,3 +72,7 @@
 .mailtoIcon{
   font-size: smaller;
 }
+
+.extraInfoRows {
```
MImran2002: If we use <strong> we can get rid of this.

MImran2002 (changes requested):
Part 1 of my review, will do more tomorrow.

app/templates/snips/studentHistoryModal.html
```diff
@@ -67,9 +67,9 @@
         {% if "Denied" in form.status.statusName %}
         <div class="leftMargin">
```
MImran2002: Image

The UI boxes are not even

app/controllers/main_routes/laborHistory.py
```diff
@@ -131,8 +131,13 @@ def populateModal(statusKey):
         currentUser = require_login()
         if not currentUser:                    # Not logged in
             return render_template('errors/403.html'), 403
-        forms = (FormHistory.select().join(LaborReleaseForm, join_type=JOIN.LEFT_OUTER)
-                            .where(FormHistory.formID == statusKey).order_by(FormHistory.createdDate.desc(), FormHistory.formHistoryID.desc()))
+        forms = (FormHistory.select()
```
MImran2002, line 135: Is there a specific reason why multiple queries were being joined? I looked over the function and most of the function are the same as in I don't see those joined models being used at all in places that were refactored.

app/templates/snips/studentHistoryModal.html
```diff
@@ -97,9 +98,9 @@
...
           <div class="row">
-            <p class="col-sm-4 pTag">Reject Reason:</p>
+            <p class="col-sm-4 pTag extraInfoRows">Reject Reason:</p>
```
MImran2002: to get rid of these classes as <strong> can be used and has been used throughout the codebase

MImran2002 (changes requested):

app/static/css/studentHistoryModal.css
JohnCox2211: got rid of it

app/templates/snips/studentHistoryModal.html
JohnCox2211: The use of the class seemed cleaner but for consistency I changed it to strong as suggested

app/templates/snips/studentHistoryModal.html
JohnCox2211: gone

app/templates/snips/studentHistoryModal.html
JohnCox2211: made spacing consistent

app/controllers/main_routes/laborHistory.py
JohnCox2211: removed extra joins

app/logic/allPendingForms.py
```diff
@@ -283,6 +283,10 @@ def checkAdjustment(allForms):
...
         if allForms.adjustedForm.fieldAdjusted == "department":
-            newDepartment = Department.get(Department.ORG==allForms.adjustedForm.newValue)
-            allForms.adjustedForm.newValue = newDepartment.DEPT_NAME
-            allForms.adjustedForm.oldValue = newDepartment.ORG + "-" + newDepartment.ACCOUNT
+            newDepartment = Department.get_or_none(Department.ORG==allForms.adjustedForm.newValue)
+            if newDepartment:
+                allForms.adjustedForm.newValue = newDepartment.DEPT_NAME
+                allForms.adjustedForm.oldValue = newDepartment.ORG + "-" + newDepartment.ACCOUNT
+            else:
+                allForms.adjustedForm.newValue = "Unknown" + " - " + allForms.adjustedForm.newValue
```
RueHaile: Each value should be validated against the Department model independently so that one failure doesn't "poison" the label of the other.

RueHaile:
Each value should be validated against the Department model independently so that one failure doesn't "poison" the label of the other. For all pendingform. py

RueHaile:
image

JohnCox2211:
> image

made a modification to the error handling here to address your concern

RueHaile:
> > image
> 
> made a modification to the error handling here to address your concern
looks good to me!

app/controllers/main_routes/laborHistory.py
RueHaile: the change made on the comment "Each value should be validated against the Department model independently so that one failure doesn't "poison" the label of the other. For all pendingform. py" looks good to me.

RueHaile (approved):
looks good to me

app/templates/snips/studentHistoryModal.html
MImran2002: image

This still exist can we meet and go over so that it is not an inconsistency from my environment.

MImran2002 (changes requested):

MImran2002 (approved):

Issue #534, opened 2025-09-12
'Manage' modal for Overloads is erratic after sorting.


BrianRamsay:
After sorting the Pending Overloads table (from the Pending Forms page), clicking on "Manage" from the Actions menu results in erratic behavior. Sometimes the modal only shows up after the second attempt, sometimes it does not show up at all. It appears to work normally before sorting.

BrianRamsay:
![](https://github.trello.services/images/mini-trello-icon.png) [Manage modal for Overloads is erratic after sorting](https://trello.com/c/9iDFtPjM/1005-manage-modal-for-overloads-is-erratic-after-sorting)

PR #564, opened 2026-02-02
Erratic loads

RueHaile:
Fixes issue #534 

Changes:
The erratic nature of the display for the overlode pending forms and pending release forms. One of the main bugs was the use of preloaded data in the pop-up module when displaying a new request. After a buffer time, the modal would be repopulated with the new request. The other problem was the lack of padding due to the HTML code layering, which required the user to press directly on the text to grant access.

Testing:
- Navigate to pending forms under admin 
- Pick either pending forms or pending release forms
- Attempt to click either manage or view notes following the action button.
image

### DIFF
--- a/app/static/css/allPendingForms.css
+++ b/app/static/css/allPendingForms.css
@@ -112,10 +112,7 @@ a.hover_indicator:active {
   height: 50px;
   border: 5px solid #f3f3f3; 
   border-radius: 50%;
- 
-  border-top: 5px solid #3498db; 
-  
-  
+  border-top: 5px solid #3498db;
   animation: spin 1s linear infinite; 
 }
 
--- a/app/static/css/allPendingForms.css
+++ b/app/static/css/allPendingForms.css
@@ -107,3 +107,24 @@ a.hover_indicator:active {
   text-align: center !important;
 }
 
+.spinner {
+  width: 50px;
+  height: 50px;
+  border: 5px solid #f3f3f3; 
+  border-radius: 50%;
+ 
+  border-top: 5px solid #3498db; 
+  
+  
+  animation: spin 1s linear infinite; 
+}
+
+
+@keyframes spin {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
--- a/app/static/js/allPendingForms.js
+++ b/app/static/js/allPendingForms.js
@@ -358,8 +358,10 @@ function loadOverloadModal(formHistoryID, laborStatusFormID) {
   This method sends an AJAX call to recieve data used to populate
   the overload modal.
   */
+
+  $("#overloadModal").find(".modal-content").html('<div class="modal-body">Loading...</div>');
+  $("#overloadModal").find('.modal-content').load('/admin/overloadModal/' + formHistoryID) ;
   $("#overloadModal").modal("show");
-  $("#overloadModal").find('.modal-content').load('/admin/overloadModal/' + formHistoryID);
 }
 
 
--- a/app/static/js/allPendingForms.js
+++ b/app/static/js/allPendingForms.js
@@ -579,20 +579,20 @@ function submitRelease(formHistoryID) {
 
 
 function toggleNotesLog(laborStatusFormID, formHistoryID) {
-  var $logDiv = $('#logNotesDiv_' + formHistoryID); 
-  var $btn = $('#modalNote_' + formHistoryID);
-
-  if ($logDiv.css('display') === 'none') {
-    $btn.html('Hide Notes');
-    $logDiv.css('display', 'block');
-    getNotes(laborStatusFormID, formHistoryID); 
+  /*
+  This method toggles the 'Notes' log at the bottom of the
+  'Overload' and 'Release' modal to show/hide it
+  */
+  if ($('.logNotesDiv').css('display') == 'none') {
+    var modalViewNotesID = '#modalNote_' + String(formHistoryID)
+    $(modalViewNotesID).html('Hide Notes')
+    getNotes(laborStatusFormID)
+    $('.logNotesDiv').css('display', 'block')
   } else {
-    $btn.html('Notes');
-    $logDiv.css('display', 'none');
-    notesCounter(laborStatusFormID, formHistoryID);
+    notesCounter(laborStatusFormID, formHistoryID)
+    $('.logNotesDiv').css('display', 'none')
   }
 }
-
 
 function notesCounter(laborStatusFormID, formHistoryID){
   /*
--- a/app/static/js/allPendingForms.js
+++ b/app/static/js/allPendingForms.js
@@ -353,10 +353,9 @@ function clearTextArea() { //makes sure that it empties text areas and p tags wh
 
 
 
-function loadOverloadModal(formHistoryID, laborStatusFormID) {
+function loadOverloadModal(formHistoryID) {
   /*
-  This method sends an AJAX call to recieve data used to populate
-  the overload modal.
+  This method shows the modal and renders is loading when it is loading
   */
 
   $("#overloadModal").find(".modal-content").html('<div class="modal-body">Loading...</div>');
@@ -365,7 +364,7 @@ function loadOverloadModal(formHistoryID, laborStatusFormID) {
 }
 
 
-function loadReleaseModal(formHistoryID, laborStatusFormID) {
+function loadReleaseModal(formHistoryID) {
   $("#modalRelease").find('.modal-content').html('<div class="modal-body">Loading...</div>');
   $("#modalRelease").find('.modal-content').load('/admin/releaseModal/' + formHistoryID);
   $("#modalRelease").modal("show");
--- a/app/static/js/allPendingForms.js
+++ b/app/static/js/allPendingForms.js
@@ -594,35 +594,26 @@ function toggleNotesLog(laborStatusFormID, formHistoryID) {
   }
 }
 
-function notesCounter(laborStatusFormID, formHistoryID) {
+function notesCounter(laborStatusFormID, formHistoryID){
+  /*
+  This method displays the number of admin notes a Labor
+  Status Form has
+  */
   var data = {'laborStatusFormID': laborStatusFormID}
   data = JSON.stringify(data)
-
-  // lock the Notes span 
-  var viewNotesID = '#notes_' + String(formHistoryID)
-  $(viewNotesID).css('pointer-events', 'none')
-  $(viewNotesID).css('opacity', '0.5')
-
   $.ajax({
     method: "POST",
     url: '/admin/notesCounter',
     data: data,
     contentType: 'application/json',
     success: function(response) {
+      var viewNotesID = '#notes_' + String(formHistoryID)
       var modalViewNotesID = '#modalNote_' + String(formHistoryID)
-
-      // update buttons with the count
       $(viewNotesID).html('Notes (' + response['noteTotal'] + ')')
       $(modalViewNotesID).html('View Notes (' + response['noteTotal'] + ')')
     },
-    error: function(request, status, error) {
+    error: function(request,status,error){
       console.log(request.responseText);
-    },
-    complete: function() {
-      setTimeout(function() {          
-        $(viewNotesID).css('pointer-events', '')
-        $(viewNotesID).css('opacity', '')
-      }, 900);
     }
   });
 }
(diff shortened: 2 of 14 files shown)


Meatchema:
- The first time I tested by just going to overload opening and pressing manage. When I did this the modal showed up and processed well.
- The second time, after sorting by "created" and then clicking manage the page whited out (went blank).

image

- After reloading the blank page, it went back to "normal", but it did not bring up the modal and was no longer sorted by "created".

image

- Next, I tried sorting and then choosing "notes" (it took two presses of the notes button to work each time I tested it). The model came up and submitted properly but then it refreshed the page and undid the sorting. 
- Specifically for the pending release tab, the manage button needs to be clicked twice as well and suffers from gradual lag (this is when the screen gets darker but doesn'tm show modal). The page also resets and undo's the sorting.
image

Overall: The modal seems to use the right student information and always seems to show up even after multiple clicks. But it still has the buffer time issue. Also is this "The other problem was the lack of padding due to the HTML code layering, which required the user to press directly on the text to grant access." talking about the functionality of the buttons for manage and notes or something else?

RueHaile:
> * The first time I tested by just going to overload opening and pressing manage. When I did this the modal showed up and processed well.
> * The second time, after sorting by "created" and then clicking manage the page whited out (went blank).
> 
> image
> * After reloading the blank page, it went back to "normal", but it did not bring up the modal and was no longer sorted by "created".
> 
> image
> * Next, I tried sorting and then choosing "notes" (it took two presses of the notes button to work each time I tested it). The model came up and submitted properly but then it refreshed the page and undid the sorting.
> * Specifically for the pending release tab, the manage button needs to be clicked twice as well and suffers from gradual lag (this is when the screen gets darker but doesn'tm show modal). The page also resets and undo's the sorting.
> 
> image
> Overall: The modal seems to use the right student information and always seems to show up even after multiple clicks. But it still has the buffer time issue. Also is this "The other problem was the lack of padding due to the HTML code layering, which required the user to press directly on the text to grant access." talking about the functionality of the buttons for manage and notes or something else?

Done!

Meatchema:
image

The modals both in manage and notes now show up with ease and no longer refresh the page which is good. I did notice however that the modal wouldn't open up at all for the notes that has zero notes. This logically makes sense, but I was wondering if it would be a good idea for the option to disappear if there is less than 1 note available? Because it seems a little odd to still be able to click it, but nothing happens, meaning that the user won't get the signal that the button tried to work.

image

RueHaile:
Thanks for the review. It looks like it is also a way to enter notes, not just to view them, and this seems intentional. I will talk to Brian and see if it is a bug or a feature.

RueHaile:
> Thanks for the review. It looks like it is also a way to enter notes, not just to view them, and this seems intentional. I will talk to Brian and see if it is a bug or a feature.

I have spoken to Brian, and he said it is also used as an input space, so a more generic wording is appropriate. I have done that. It is no longer "view notes" but merely "notes". There is a question of whether people really use this feature, but we can't just remove it. It will be addressed in future pr's.

MImran2002 (changes requested):
Fix 1: The db_test.py should be uncommitted even though it fixes the Tracy issue I remember talking with Brian about this and not moving forward with those changes.

Fix 2: Update the comment on this function app>static>js>allPendingForms.js>loadOverloadModal because there is no AJAX call and one of the parameter is not in use.

Fix 3: Remove unused parameters or variables like laborStatusFormID in loadReleaseModal.

Fix 4: in allPendingForms.html around line 209 and line 218, calls the loadoverloadModal but the second parameter is never used in the js function so having the parameter inputted in the html side is in valid cause the parameter is never used.

RueHaile:
> Fix 1: The db_test.py should be uncommitted even though it fixes the Tracy issue I remember talking with Brian about this and not moving forward with those changes.
> 
> Fix 2: Update the comment on this function app>static>js>allPendingForms.js>loadOverloadModal because there is no AJAX call and one of the parameter is not in use.
> 
> Fix 3: Remove unused parameters or variables like laborStatusFormID in loadReleaseModal.
> 
> Fix 4: in allPendingForms.html around line 209 and line 218, calls the loadoverloadModal but the second parameter is never used in the js function so having the parameter inputted in the html side is in valid cause the parameter is never used.

done

Meatchema:
Both buttons seem to work as intended.
image
However, the filters seem to still be resetting after reloading which makes sense, but I am not sure if that is a concern. Is the concern mainly the buttons functionality, the smoothness in which they function, or how it effects filters? 

My answers:

- Both buttons show up and work properly. The correct information shows and I don't have to click directly on the letters to make it work.
- They both have increase in smoothness of usage (they don't require multiple clicks anymore, though there is still a little bit of a load buffer)
- If you use a filter and then make a note or click and use manage the page will reset making it harder to go find the persons information you just edited to see your changes.

Note:
Is there something specific I should be looking for in my review?

Meatchema (commented):
Reposting this comment since the other didn't seem to be flagged as my review:
Both buttons seem to work as intended.
image
However, the filters seem to still be resetting after reloading which makes sense, but I am not sure if that is a concern. Is the concern mainly the buttons functionality, the smoothness in which they function, or how it effects filters?

My answers:

Both buttons show up and work properly. The correct information shows and I don't have to click directly on the letters to make it work.
They both have increase in smoothness of usage (they don't require multiple clicks anymore, though there is still a little bit of a load buffer)
If you use a filter and then make a note or click and use manage the page will reset making it harder to go find the persons information you just edited to see your changes.
Note:
Is there something specific I should be looking for in my review?

RueHaile:
> Reposting this comment since the other didn't seem to be flagged as my review: Both buttons seem to work as intended. image However, the filters seem to still be resetting after reloading which makes sense, but I am not sure if that is a concern. Is the concern mainly the buttons functionality, the smoothness in which they function, or how it effects filters?
> 
> My answers:
> 
> Both buttons show up and work properly. The correct information shows and I don't have to click directly on the letters to make it work. They both have increase in smoothness of usage (they don't require multiple clicks anymore, though there is still a little bit of a load buffer) If you use a filter and then make a note or click and use manage the page will reset making it harder to go find the persons information you just edited to see your changes. Note: Is there something specific I should be looking for in my review?


I had spoken to brian and the revert when the page reloads is not a bug. I appreciate your comment. No, that is all thanks.

app/static/js/allPendingForms.js
```diff
@@ -353,21 +353,21 @@ function clearTextArea() { //makes sure that it empties text areas and p tags wh
...
+  $("#overloadModal").find('.modal-content').load('/admin/overloadModal/' + formHistoryID) ;
   $("#overloadModal").modal("show");
-  $("#overloadModal").find('.modal-content').load('/admin/overloadModal/' + formHistoryID);
 }
 
 
-function loadReleaseModal(formHistoryID, laborStatusFormID) {
-  $("#modalRelease").modal("show");
+function loadReleaseModal(formHistoryID) {
+  $("#modalRelease").find('.modal-content').html('<div class="modal-body">Loading...</div>');
```
ojmakinde: Why are we setting the HTML to Loading if we're going to set it to the actual content in the next line? Also I love this fix!

app/static/js/allPendingForms.js
```diff
@@ -353,21 +353,21 @@ function clearTextArea() { //makes sure that it empties text areas and p tags wh
 
 
 
-function loadOverloadModal(formHistoryID, laborStatusFormID) {
+function loadOverloadModal(formHistoryID) {
   /*
-  This method sends an AJAX call to recieve data used to populate
-  the overload modal.
+  This method shows the modal and renders is loading when it is loading
```
ojmakinde: Can you modify the docstring?

app/static/js/allPendingForms.js
```diff
@@ -353,21 +353,21 @@ function clearTextArea() { //makes sure that it empties text areas and p tags wh
...
   */
+
+  $("#overloadModal").find(".modal-content").html('<div class="modal-body">Loading...</div>');
```
ojmakinde: Also curious about the use of the "Loading..." text here

ojmakinde (commented):
Looks great. Just a few small changes and should be ready to merge. Great work.

app/static/js/allPendingForms.js
RueHaile: Good one, I have fixed it now. Currently, it will no longer be in that order; it will set it to loading and then render it, giving it time to display the content from the ajax request.

app/static/js/allPendingForms.js
RueHaile: The need for the "Loading..." is because the modal content persists from the last time it was opened. This creates a "flash" of the wrong student's data, followed by a self-correction when it is updated.

Through the way that it is handeld we are creating a buffer that is user-friendly (as it gives visual feedback), at the same time, renders the right information.

app/static/js/allPendingForms.js
RueHaile: Yes, done.

ojmakinde (commented):
LGTM. @BrianRamsay, apparently, the spinner that used to exist on the pending forms modal no longer exists. Did you get rid of it?? Also, we're using a simple "Loading..." text for that transitional state now. Thoughts?

RueHaile:
> LGTM. @BrianRamsay, apparently, the spinner that used to exist on the pending forms modal no longer exists. Did you get rid of it?? Also, we're using a simple "Loading..." text for that transitional state now. Thoughts?

I have set the spinner instead of the "Loading..."  so it is back; however, I only added it, and I couldn't find its implementation anywhere else.

JohnCox2211 (changes requested):
When reviewing this PR, I was still getting the occasional issue of clicking notes and notes not opening up on all of the Pending Forms pages. The page would scroll to the top like it does when the notes modal opens, but the modal would not appear. This would occur once every 3-10 times or so. 

I also noticed that the deny button sometimes still doesn't show the modal, every 3-10 tries, much like with the notes button. However, instead of the page scrolling to the top, nothing would happen except the dropdown with the buttons closing. 

I do not get the same problems when clicking modify or manage. 

Though the issue specifies the issue is with the manage modal, if it happens with all of the other buttons with modals in the same button dropdowns, maybe those fixes could be included?

RueHaile:
> When reviewing this PR, I was still getting the occasional issue of clicking notes and notes not opening up on all of the Pending Forms pages. The page would scroll to the top like it does when the notes modal opens, but the modal would not appear. This would occur once every 3-10 times or so.
> 
> I also noticed that the deny button sometimes still doesn't show the modal, every 3-10 tries, much like with the notes button. However, instead of the page scrolling to the top, nothing would happen except the dropdown with the buttons closing.
> 
> I do not get the same problems when clicking modify or manage.
> 
> Though the issue specifies the issue is with the manage modal, if it happens with all of the other buttons with modals in the same button dropdowns, maybe those fixes could be included?

thank you!. I have fixed the note's erratic nature; however, the "deny" is not part of this PR.

JohnCox2211 (changes requested):
I am still seeing the previous issue where the notes button does not always open the notes modal. As before, once every 5-10 tries on different notes button presses, it will not open and will instead scroll the page to the top.

RueHaile:
> I am still seeing the previous issue where the notes button does not always open the notes modal. As before, once every 5-10 tries on different notes button presses, it will not open and will instead scroll the page to the top.

I discussed with @BrianRamsay have resolved the issue by disabling the button, creating a time delay that would prohibit the null response. 

Previously when you clicked Actions, notesCounter() would fired and went to the server. If you clicked Notes before it came back, the Notes request would fire but then notesCounter would return and wipe the button HTML; killing the Notes request and preventing the modal from opening.

JohnCox2211 (approved):
Looks good to me, all comments have been addressed.

RueHaile:
> LGTM. @BrianRamsay, apparently, the spinner that used to exist on the pending forms modal no longer exists. Did you get rid of it?? Also, we're using a simple "Loading..." text for that transitional state now. Thoughts?

I have found it it is .loader and it is in formHistory.css however it would require a more intensive change and DOM manupliation  as currently the .however, if it d

app/static/css/allPendingForms.css
```diff
@@ -107,3 +107,24 @@ a.hover_indicator:active {
   text-align: center !important;
 }
 
+.spinner {
+  width: 50px;
+  height: 50px;
+  border: 5px solid #f3f3f3;
```
MImran2002, line 113: big space; to delete lines not used

app/static/js/allPendingForms.js
```diff
@@ -252,8 +252,8 @@ function resendApprovalLink(formId) {
   });
 }
 
-function getNotes(formId) {
-  $.ajax({
+function getNotes(formId, formHistoryID) {
+    $.ajax({
```
MImran2002: identation

MImran2002 (approved):
looks good

PR #567, opened 2026-03-16
Removed evaluation column from term management page

### LINKED ISSUE #522: Remove Evaluation buttons from Term Management
We no longer need the evaluation buttons on term management. We should get rid of the whole column

Image

Jonimz:
Fixes #522

Changes: 
Removed the Evaluations (Midyear | Final) column from the Term Management table as it is no longer needed. This includes the column header and the corresponding table cell containing the midyear and final evaluation toggle buttons for each term row.

Testing:
* Navigate to Term Management under admin
* Verify the Evaluations column no longer appears in the table
* Verify the table layout is not broken after column removal

Screenshot 2026-03-16 at 2 50 06 PM


JohnCox2211 (changes requested):
The column for the midyear and final evaluations is gone, but connected code of the backend is still there. I don't think we use it anywhere else, so it should be removed too it seems in files like termManagement.py, term.py, studentLaborEvaluations.py, studentLaborEvaluations.html. 

All of those files have references to variables used by this functionality like isFinalEvaluationOpen or isMidyearEvaluationOpen.

JohnCox2211:
Error upon trying to navigate to the manage terms page.

image

There are still references to evaluations like isFinalEvaluationOpen being part of render_template in studentLaborEvaluation.py, and is referenced in studentLaborEvaluation.html.

JohnCox2211 (changes requested):
I am still getting the previous peewee error when on the backup database (it is not present when using test)

Image

MImran2002 (changes requested):
There are multiple .DS_Store being pushed they shouldn't be there along with extensions and fetch head. And this is anothe rtime it happened, it happened with me too I think we should add this in git ignore

MImran2002:
talked with Brian, we will be keeping the model data to ensure the data is still there, most of the changes are to aim at UI and front-end access and likely backend access too,

app/controllers/main_routes/main_routes.py
```diff
@@ -80,9 +80,8 @@ def downloadSupervisorPortalResults():
         formSearchResults.searchType,
         requestedLSFs=formHistories, 
         additionalSpreadsheetFields=[],
-        includeEvals=False
     )
-    return send_file(excel.relativePath, as_attachment=True, attachment_filename=excel.relativePath.split('/').pop())
+    return send_file(excel.relativePath, as_attachment=True, download_name=excel.relativePath.split('/').pop())
```
bakobagassas: is there a reason why we are changing attachment_filename to download_file name ?

app/controllers/admin_routes/termManagement.py
```diff
@@ -115,22 +115,3 @@ def termStatusCheck():
     except Exception as e:
         print(e)
         return jsonify({"Success": False})
-
```
bakobagassas: Why are we deleting the manageEval function? I thought we were only getting rid of the current front end and not the actual manage evaluation logic ?

bakobagassas: Never mind, we are getting rid of the backend too! so it's good!

app/controllers/main_routes/studentLaborEvaluation.py
```diff
@@ -115,48 +93,23 @@ def sle(statusKey):
...
-            is_midyear_evaluation = False
-        else:                                                               # Midyear
-            is_midyear_evaluation = True
         try:
-            sle = StudentLaborEvaluation.get(formHistoryID = laborHistoryForm, is_submitted = False, is_midyear_evaluation = is_midyear_evaluation)
+            sle = StudentLaborEvaluation.get(formHistoryID = laborHistoryForm, is_submitted = False)
             sle.delete_instance()
         except DoesNotExist:
             pass
-        # Then, save the new record
```
bakobagassas: Let's keep the existing comments, unless the code does something else

app/templates/main/studentLaborEvaluation.html
```diff
@@ -497,19 +445,12 @@ <h3>{{ form.transcriptComments.label }}
         </p>
 
       </div>
-      {% if not existing_final_evaluation and (not existing_midyear_evaluation or isFinalEvaluationOpen) %}
+      {% if not existing_evaluation %}
```
bakobagassas: - why are we converting existing_final_evaluation to existing evaluation if we are getting rid of evaluations? 
- In the issue, are we keeping existing evaluation for any purpose?

app/controllers/main_routes/main_routes.py
Jonimz: I don’t believe this change was part of the evaluation work. It may have come in from one of the merges with `development`. I’ll check the commit history to confirm before changing it.

app/controllers/main_routes/studentLaborEvaluation.py
Jonimz: Fixed!

BrianRamsay (approved):

Issue #520, opened 2025-08-21
Fix Labor History form ordering


Karina-Agliullova:
Labor History form ordering is incorrect for forms that are in the same term. For example, if a student has been released from a position and then takes a new position the ordering should match the rest of the history, with the newest position at the top.

BrianRamsay:
![](https://github.trello.services/images/mini-trello-icon.png) [Labor History form ordering is incorrect for forms that are in the same term.](https://trello.com/c/prjE0RME/897-labor-history-form-ordering-is-incorrect-for-forms-that-are-in-the-same-term)

PR #570, opened 2026-03-26
Labor History Form Ordering

JohnCox2211:
Fixes issue #520 

### Changes:
Corrected the order of student labor status forms to match the order in which the student has started the labor positions rather than going by term. 

### Testing:

- Navigate to the Student Search tab 
- Search for any student such as John Cox and navigate to the Labor History page
- The order of the forms on this page should now be ordered based on when the student started the position

**My labor history page before**:
image

**My labor history page after**:

image

### DIFF
--- a/app/controllers/main_routes/laborHistory.py
+++ b/app/controllers/main_routes/laborHistory.py
@@ -57,26 +57,41 @@ def laborhistory(id):
                 if len(authorizedForms) == 0:
                     return render_template('errors/403.html'), 403
         
-        authorizedForms = Term.order_by_term(list(authorizedForms.objects()), reverse=True)
-        downloadId = saveFormSearchResult("Labor History", authorizedForms, "studentHistory")
+        ###########################################
+        authorizedForms = list(authorizedForms.objects())
 
         laborStatusFormList = ','.join([str(form.formID.laborStatusFormID) for form in studentForms])
-        # modify status display for overload and release forms
+
         formIds = [form.formID for form in authorizedForms]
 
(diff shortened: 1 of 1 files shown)


app/controllers/main_routes/laborHistory.py
```diff
@@ -56,8 +56,9 @@ def laborhistory(id):
 
                 if len(authorizedForms) == 0:
                     return render_template('errors/403.html'), 403
-        
-        authorizedForms = Term.order_by_term(list(authorizedForms.objects()), reverse=True)
+
+        authorizedForms = list(authorizedForms.objects())
+        authorizedForms.sort(key=lambda f: f.startDate or datetime.date.min, reverse=True)
```
Arohasina, line 61: Just to confirm, should labor forms without a start date appear at the bottom (current behavior with datetime.date.min) or at the top (would require datetime.date.max)? This is a good fallback, so I doubt this will matter if we do max or min. 

Other than that, everything looks good to me.

Arohasina (changes requested):

app/controllers/main_routes/laborHistory.py
JohnCox2211: Since min is the earliest possible date, it will appear at the bottom of the list if there is no attached date.

Meatchema (approved):
Both the code and the front end look well done. I see nothing else to change for Labor History Form ordering. Newly added positions also automatically go into the right order.

Arohasina (approved):
looks good

Issue #526, opened 2025-08-22
Reconcile contrasting confirmation messages


bakobagassas:
When creating a new form in LSF, there is a message like this that shows up to notify that the creation was not successful and well as a flash message showing failure. <img width="1845" height="884" alt="Image" src="https://github.com/user-attachments/assets/23aad8f2-0813-4313-8275-69f2676f4f18" />

Then like one second later, we see a flash message like this saying that the creation was successful 
<img width="1787" height="768" alt="Image" src="https://github.com/user-attachments/assets/0871c836-d53a-4c29-bae0-6b423abb7145" />

BrianRamsay:
In general we need to solidify how we report success and failure when creating a form. In addition to the above behavior where there's an error reported but a successful flash confirmation, it's possible for the request to fail because of email sending and a negative flash message, but the form will still be created.

BrianRamsay:
![](https://github.trello.services/images/mini-trello-icon.png) [Refactor inconsistent LSF submission responses](https://trello.com/c/aNmkpEd3/1000-refactor-inconsistent-lsf-submission-responses)

PR #571, opened 2026-04-02
526 contrasting confirmation messages

Kafui123:
Fixes issue https://github.com/BCStudentSoftwareDevTeam/lsf/issues/526

Changes:
- Fixed the issue where an error message was shown, but a successful flash confirmation still appeared afterward. Now the feedback is consistent... no mixed signals for the user. Flash message errors are considered and both back-end error and front-end error UI are done fro both developers and users.
- Adjusted the logic so that success and failure states are clearly separated and reported correctly, instead of overlapping or triggering both paths.
- Addressed the case where email sending could fail, but the form would still be created, which previously caused a misleading failure message even though the form existed.
- Ensured that form creation and email sending happen within a single transaction block.
- Front-end flash message is more responsive, stacked

Testing:

- Click on administration and then manage terms. 
- Select a valid term under any academic year. 
- Create a new labor status form and then click submit
- If everything succeeded, flash success message would appear 
- if you want to break it and see an error flash, you can remove a parameter for one of those functions that send an email to see the error: createOverloadFormAndFormHistory , emailDuringBreak(checkForSecondLSFBreak 
- If the form got created, but no email was sent, the form would not get created and an error flash will appear.

Screenshot 2026-04-02 145031
Screenshot 2026-04-02 145042


bakobagassas (changes requested):
Why are we changing 6 test files ? I don't think most of those changes are necessary

Kafui123:
> Why are we changing 6 test files ? I don't think most of those changes are necessary

I had to make those changes because they(the test files) were failing as a result of the new changes I added.

app/logic/emailHandler.py
```diff
@@ -16,7 +16,7 @@
 from app import app
 import os
 from datetime import datetime, date
-
+from app.logic.utils import makeThirdPartyLink
```
bakobagassas: makeThirdPartyLink is imported but not used in the code.

app/static/js/laborStatusForm.js
Arohasina, line 1: line 766: Missing error handler for AJAX request. If the network fails or the server returns an error (form creation fails), the user gets no feedback. Add an error callback to handle failures.

tests/code/test_adminManagement.py
```diff
@@ -5,42 +5,54 @@
...
     # Test adding saas admin
-    addAdmin(user, 'saas')
+    addAdmin(user, 'Saas')
     assert user.isSaasAdmin
+    
+    # cleanup
+    user.isLaborAdmin = False
+    user.isFinancialAidAdmin = False
+    user.isSaasAdmin = False
+    user.save()
```
bakobagassas: It would be simpler to use maindb.transaction as transaction: transaction.rollback for clean up

tests/code/test_adminManagement.py
```diff
@@ -5,42 +5,54 @@
 
 @pytest.mark.integration
 def test_addAdmin():
-    newAdmin = "pearcej"
+    newAdmin = "jonesj"
```
bakobagassas: Why are we changing this? It seems as if we have been using pearcej in other places, I think we should just keep that.

tests/code/test_adminManagement.py
```diff
@@ -5,42 +5,54 @@
...
     # Test adding financial aid admin
-    addAdmin(user, 'finAid')
+    addAdmin(user, 'FinancialAid')
+    
     assert user.isFinancialAidAdmin
 
     assert not user.isSaasAdmin
     # Test adding saas admin
-    addAdmin(user, 'saas')
+    addAdmin(user, 'Saas')
```
bakobagassas: Why are we capitalizing some parameters?

app/static/js/laborStatusForm.js
```diff
@@ -192,98 +192,98 @@ function updateDate(obj) { // updates max and min dates of the datepickers as th
...
+    $(list).each(function (i, hours) {
       selectedHoursPerWeek.append($("<option />").text(hours));
     });
   }
-    if (fillhours == ""){
-      $(".selectpicker").selectpicker("refresh");
-    }
+  if (fillhours == "") {
+    $(".selectpicker").selectpicker("refresh");
   }
```
bakobagassas, line 286: These changes are not needed (they are space changes in too many places)

Arohasina (changes requested):

bakobagassas (changes requested):

app/logic/statusFormFunctions.py
```diff
@@ -57,41 +57,43 @@ def createOverloadFormAndFormHistory(rspFunctional, lsf, creatorID, host=None):
...
-        email.laborStatusFormSubmitted()
+        formHistory = FormHistory.create( formID = lsf.laborStatusFormID,
+                            historyType = "Labor Status Form",
+                            overloadForm = None,
+                            createdBy   = creatorID,
+                            createdDate = date.today(),
+                            status      = "Pre-Student Approval")
 
-    return formHistory
+        if not formHistory.formID.termCode.isBreak and not isOverload:
```
Arohasina: Is there a reason why email.laborStatusFormSubmitted() has been removed? Do we not want the email to be sent? Also, this should have error handling in case email sending fails, perhaps

Arohasina (changes requested):

app/static/js/laborStatusForm.js
Arohasina: also how do we handle if email sending fails?

Arohasina:
Your tests now handle success cases, but could you consider adding an integration test for partial/full submission failure scenarios after you have added the fixes mentioned above?

app/logic/statusFormFunctions.py
```diff
@@ -57,41 +57,43 @@ def createOverloadFormAndFormHistory(rspFunctional, lsf, creatorID, host=None):
...
+                                                createdBy   = creatorID,
+                                                createdDate = date.today(),
+                                                status      = "Pre-Student Approval")
+            email = emailHandler(formOverload.formHistoryID)
+            link = makeThirdPartyLink("student", host, formOverload.formHistoryID)
+            email.LaborOverLoadFormSubmitted(link)
 
-    if not formHistory.formID.termCode.isBreak and not isOverload:
-        email = emailHandler(formHistory.formHistoryID)
-        email.laborStatusFormSubmitted()
```
Arohasina, line 91: Is there a reason why email.laborStatusFormSubmitted() has been removed? Do we not want the email to be sent? Also, this should have error handling in case email sending fails, perhaps

app/static/js/laborStatusForm.js
Kafui123: Great catch! 
Also, regarding  handling the email sending failure, I put everything inside a try and except block, so if an email sending fails, it exits the try block, moves to the except block, and appends false into the all_forms array.(marks the form submission as false)

tests/code/test_adminManagement.py
Kafui123: in the addAdmin function, it is expecting the adminType to be capitalized (i.e the first letter)

app/logic/statusFormFunctions.py
Kafui123: I  have added the email.laborStatusFormSubmitted() back. Also, these is an error handling that happens in the laborstatusform.py file.

Kafui123:
> Your tests now handle success cases, but could you consider adding an integration test for partial/full submission failure scenarios after you have added the fixes mentioned above?



> Your tests now handle success cases, but could you consider adding an integration test for partial/full submission failure scenarios after you have added the fixes mentioned above?



> Your tests now handle success cases, but could you consider adding an integration test for partial/full submission failure scenarios after you have added the fixes mentioned above?



Since the main change in this pr was to put everything inside a mainDB.atomic block so that a form would not get submitted if email sending fails i do not think that it would be necessary testing the mainDB.atomic() block since that is peewee's behavior.

Arohasina (approved):
looks good !

app/logic/statusFormFunctions.py
```diff
@@ -57,41 +57,43 @@ def createOverloadFormAndFormHistory(rspFunctional, lsf, creatorID, host=None):
...
+        formHistory = FormHistory.create( formID = lsf.laborStatusFormID,
+                            historyType = "Labor Status Form",
+                            overloadForm = None,
+                            createdBy   = creatorID,
```
MImran2002, line 86: shouldn't the email.laborStatusFormSubmitted be here?

app/models/__init__.py
```diff
@@ -1,8 +1,9 @@
+from app import app
```
MImran2002: Can I know the rationale for this as working on init.py file affect most of the backend

MImran2002 (changes requested):

app/models/__init__.py
Kafui123: I added these lines to resolve a circular import issue where other files depended on the db instance from this file. At the time, those modules were being imported before the database had been properly initialized, which led to errors when accessing db. Importing the Flask app and initializing db with it ensured that the database instance was available when those modules were loaded.

app/logic/statusFormFunctions.py
Kafui123: Yes, the call is intentionally handled later in the flow. For overload forms, the email is triggered earlier on line 81 via email.LaborOverloadFormSubmitted(link)  and for regular status forms, the emaila gets sent on line 91 after the FormHistory record is created.

MImran2002 (commented):
there are whitespaces, some changes I will revert to reduce the code change scope that are not related to the issue description

app/models/__init__.py
MImran2002: That is valid, there some local changes that I have to make to have the system up and running. But in those cases where usually it's the environment file unless explicitly stated by the issue i would keep it in the local change only instead of pushing it. So I am gonna go and revert it as pushing this through would not be compatible with other environment builds for other people. However, if its something that persists in your local change definitely save it on your end cause I have been in your situation and I saved mine in like a google doc.

app/logic/statusFormFunctions.py
MImran2002: You have not created the laboroverloadformsubmitted yet so will be doing that

app/logic/statusFormFunctions.py
```diff
@@ -223,13 +225,18 @@ def emailDuringBreak(secondLSFBreak, term):
...
-            email.notifyAdditionalLaborStatusFormSubmittedForBreak()
+    try: 
+        if term.isBreak:
+            isOneLSF = json.loads(secondLSFBreak)
+            formHistory = FormHistory.get(FormHistory.formHistoryID == isOneLSF['formHistoryID'])
+            email = emailHandler(formHistory.formHistoryID)
+            email.laborStatusFormSubmitted()
+            if(len(isOneLSF["previousSupervisorNames"]) > 1): #Student has more than one lsf. Send email to both supervisors and student
+                email.notifyAdditionalLaborStatusFormSubmittedForBreak()
+    except Exception as e:
```
MImran2002, line 236: is the error handling because the contrasting message is removed?

app/static/js/laborStatusForm.js
```diff
@@ -724,159 +725,154 @@ function createModalContent() { // Populates Submit Modal with Student informati
...
-           failed_students.forEach(function(item){
-              createAndFillTable(item);
-           });
-           $('#SubmitModal').modal('hide');
-           display_failed=[];
-         }
+  $.ajax({
+    method: "POST",
+    url: "/laborstatusform/userInsert",
+    data: JSON.stringify(globalArrayOfStudents),
```
MImran2002: will be reverting this file tomorrow as whitespace even though it can be good for readability won't be needed here as it delays the review process

app/logic/statusFormFunctions.py
MImran2002: The rationale that email.laborStatusFormSubmitted() should be added again is because if the userinsert is done and its called the creation of a labor status form being created through createoverloadFormAndFormhisotry there is no way to send out the email unless it is overload

tests/code/test_tracy.py
```diff
@@ -15,132 +16,146 @@ def test_init(self, tracy):
 
     @pytest.mark.integration
     def test_getStudents(self, tracy):
-        students = tracy.getStudents()
-        assert ['Elaheh','Guillermo','Jeremiah','Kat', 'Oluwagbayi', 'Test', 'Tyler'] == [s.FIRST_NAME for s in students]
-        assert ['718','300','420','420', '883', '700', '420'] == [s.STU_CPO for s in students]
+        with app.app_context():
+            print("HERE")
```
fritzj2: Remove this print statement

tests/code/test_tracy.py
```diff
@@ -15,132 +16,146 @@ def test_init(self, tracy):
...
+            students = tracy.getStudents()
+            print("THERE")
```
fritzj2: This too

app/controllers/main_routes/laborStatusForm.py
```diff
@@ -73,31 +74,34 @@ def userInsert():
...
-            createOverloadFormAndFormHistory(rspFunctional[i], lsf, currentUser, host=request.host) 
-            try:
-                emailDuringBreak(checkForSecondLSFBreak(term.termCode, student.ID), term)
-            except Exception as e:
-                print("Error when sending emails during break: " + str(e))
-
+            with mainDB.atomic():
+                lsf = createLaborStatusForm(student, supervisor.ID, department.departmentID, term, rspFunctional[i])
+                try:
+                    createOverloadFormAndFormHistory(rspFunctional[i], lsf, currentUser, host=request.host)
```
fritzj2: The Overload form should be outside of the second try statement.
The second try should be just for email errors.

fritzj2 (changes requested):

app/controllers/main_routes/laborStatusForm.py
Kafui123: great catch

tests/code/test_apiEndpoint.py
```diff
@@ -29,14 +30,14 @@ def test_getLaborInformation():
                                isLaborAdmin = None,
                                isFinancialAidAdmin = None,
                                isSaasAdmin = None)
-        
+        print("what about here")
```
fritzj2: Another Print statement

fritzj2 (changes requested):
Other than this print statement, are you all capable of creating LSF in the development environment? 
I can only create them if I turn off the email handling. But apart from that it seems to be working fine.

Kafui123:
> Other than this print statement, are you all capable of creating LSF in the development environment? I can only create them if I turn off the email handling. But apart from that it seems to be working fine.


@fritzj2 I think this might be related to the email configuration in the development environment. Since the email handling is inside the mainDB.atomic() transaction, if emailDuringBreak() fails because the app password or email credentials aren't set up in secretconfig.yaml, the exception gets re-raised and causes the transaction to roll back.

So even though the LSF creation itself seems to be working, the failed email is causing the whole transaction to fail. Could you check if the email app password is set up in your development environment? If it is, then I can take a closer look at the email handling and see what else might be causing it.

app/logic/statusFormFunctions.py
```diff
@@ -57,41 +57,43 @@ def createOverloadFormAndFormHistory(rspFunctional, lsf, creatorID, host=None):
...
+                            createdBy   = creatorID,
+                            createdDate = date.today(),
+                            status      = "Pre-Student Approval")
 
+        if not formHistory.formID.termCode.isBreak and not isOverload:
+            email = emailHandler(formHistory.formHistoryID)
+            email.laborStatusFormSubmitted()
+        return formHistory
+    except Exception as e:
+        print("Error creating overload form:", e)
```
MImran2002: this error message is not showing the full picture as the function create a formhistory and if its a overload form they also create a overloadformhistory so your error message should be able to detect whether there is a overloadform too which means your code isOverload should be outside the try and except to capture that distinction between a form and a form with overloadhistory

app/controllers/main_routes/laborStatusForm.py
```diff
@@ -73,31 +74,34 @@ def userInsert():
...
-            try:
-                emailDuringBreak(checkForSecondLSFBreak(term.termCode, student.ID), term)
-            except Exception as e:
-                print("Error when sending emails during break: " + str(e))
-
+            with mainDB.atomic():
+                lsf = createLaborStatusForm(student, supervisor.ID, department.departmentID, term, rspFunctional[i])
+                createOverloadFormAndFormHistory(rspFunctional[i], lsf, currentUser, host=request.host)
+                try: 
+                    emailDuringBreak(checkForSecondLSFBreak(term.termCode, student.ID), term)
```
MImran2002: this part worries me even if the transaction is rolled back if there is an error within this function, the emailDuringBreak will still delivered an email. So in all scenario where something breaks in the UserInsert function the email will not be delivered unless its during break.

MImran2002 (changes requested):

app/logic/statusFormFunctions.py
Kafui123: resolved.

app/logic/statusFormFunctions.py
```diff
@@ -92,7 +96,10 @@ def createOverloadFormAndFormHistory(rspFunctional, lsf, creatorID, host=None):
             email.laborStatusFormSubmitted()
         return formHistory
     except Exception as e:
-        print("Error creating overload form:", e)
+        if isOverload:
```
bakobagassas: looks good

bakobagassas (approved):
Amazing job on this! Will wait for a second review from Imran then merge it.

MImran2002 (approved):

fritzj2 (approved):
Everything seems to be working on my end, and there is no more code that jumps out at me in this. Good work!

PR #583, opened 2026-06-17
Fixed issue #531; Now  remove and discard has consistant msg flash color

### LINKED ISSUE #531: Inconsistent Remove Message Colors
### Issue
When removing a supervisor from a department in a department modal on the Manage Departments page under the Administration tab in the sidebar, The flash message is green. This is inconsistent with the red flash messages that show up elsewhere in the website when  doing a remove. 

### Testing

**The Page**
 
Image

**Click on a department name then then add a supervisor, note that the flash message when you add them is green**

Image

**Click 'Remove next to the supervisor's name, note that the flash message is also green**

Image

nahom70:
## Fixes #531 issue number

* Inconsistent message flash on removed and discarded messages

## Changes
* updated the manage terms to have a green ( success ) pop-up when closed and opened 
* Departments have success when a supervisor is added and removed.
* To be more explicit, the supervisor's name and ID are shown when removed and added
* Administrations now show success when admins are added in each of the sections ( Labor, Financial, SAAS ) 
* Email template shows success when the email is discarded and a warning when the save button is clicked when there is nothing to save/

## Testing
* Test by manipulating adding and removing, saving and discarding events, and making sure they are functional as programmed.

### DIFF
--- a/app/static/js/addSupervisorsToDepartment.js
+++ b/app/static/js/addSupervisorsToDepartment.js
@@ -1,14 +1,18 @@
-function addSupervisorToDepartment(supervisorID, departmentID, callback=() => {}) {
+function addSupervisorToDepartment(supervisorID, departmentID, supervisor_FirstName, supervisor_LastName, callback=() => {}) {
+  let supervisorfirstName = supervisor_FirstName
+  let supervisorLastName = supervisor_LastName
+  console.log("Names  ",supervisorfirstName, supervisorLastName)
+
     return $.ajax({
       method: "POST",
       url: `/supervisorPortal/addUserToDept`,
       data: {"supervisorID": supervisorID, "departmentID": departmentID},
       success: function(response) {
         if (response == "True") {
-          msgFlash("Supervisor has been added to department.", "success")
+          msgFlash(`Supervisor ${supervisorfirstName} ${supervisorLastName} (${supervisorID}) has been added to the department.`, "success")
           clearDropdowns()
         } else {
-          msgFlash("Supervisor is already a member of this department.", "warning")
+          msgFlash(`Supervisor ${supervisorfirstName} ${supervisorLastName} (${supervisorID}) is already a member of this department.`, "warning")
           clearDropdowns()
         }
         if (callback){
--- a/app/static/js/addSupervisorsToDepartment.js
+++ b/app/static/js/addSupervisorsToDepartment.js
@@ -1,7 +1,6 @@
-function addSupervisorToDepartment(supervisorID, departmentID, supervisor_FirstName, supervisor_LastName, callback=() => {}) {
-  let supervisorfirstName = supervisor_FirstName
-  let supervisorLastName = supervisor_LastName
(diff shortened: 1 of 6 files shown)


ArtemKurasov (approved):

BhushanSah (approved):
The Flash message has a consistent color now

BrianRamsay (changes requested):
I mentioned it in a comment on the issue at some point, but we shouldn't be using red for successful actions, whether it is adding or removal. So the proper fix on this issue is to make sure that we are using green for successful actions and red for failures across the board.

nahom70:
Have set successful events to green and failed events to red. There are warnings on the email template. 

Question on term management when the term status is changed to closed, should it be green since it is a successful action, or indicate red, given that the action itself is closed
image

ArtemKurasov:
Honestly, I'm currently not able to test the code, but what I would suggest is using a more neutral color for the messages (like blue). That would eliminate any possible ambiguity the messages might create. 

As far as I know, changing "success" to "info" turns green into blue if you use Bootstrap.

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -1,14 +1,18 @@
-function addSupervisorToDepartment(supervisorID, departmentID, callback=() => {}) {
+function addSupervisorToDepartment(supervisorID, departmentID, supervisor_FirstName, supervisor_LastName, callback=() => {}) {
```
MImran2002: consistent coding convention: if the variables are using a similar naming convention please stick to that in this case: supervisorLastName

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -1,14 +1,18 @@
...
+  let supervisorfirstName = supervisor_FirstName
+  let supervisorLastName = supervisor_LastName
+  console.log("Names  ",supervisorfirstName, supervisorLastName)
```
MImran2002: we don't use console.log aside from debugging as part of our codebase, please remove .

app/static/js/manageDepartments.js
```diff
@@ -87,16 +90,18 @@ function removeSupervisorFromDepartment () {
   let departmentID = $(`#${this.id}`).data('department')
   let supervisorID = $(`#${this.id}`).data('supervisor')
   let data = {"supervisorID": supervisorID, "departmentID": departmentID}
+  let supervisorfirstName = $(`#${this.id}`).parent().find('div').first().text().split(" ").slice(1, -1).join(" ")
```
MImran2002: instead of these very long multiple method line you can go to manageDepartment.js underneath showsupervisorsindepartment function populates the list of supervisor in the model you can go ahead and attached the supervisors first name and supervisor last name as data. In this way all you just need to do was call the .data(...) and get it instead of the long methods.

MImran2002 (changes requested):

MImran2002 (approved):
looks good and it works!

PR #589, opened 2026-06-22
Added more details in the email tracker. Redacts new information

### LINKED ISSUE #561: Store more detail in EmailTracker
To give us more information when debugging reported email issues, we should store more information when we send an email:

- `recipientEmails` - In addition to the existing `recipient`, we should record the list of actual email addresses the email was sent to. This should be redacted when the database is dumped for testing.
- `template` - Instead of the existing `subject` field, we should have a foreign key to the email template record, so we can have access to all of the information stored there, not just `subject`.
- `body` - We should store the actual text of the email that we have sent, which will aid in debugging incorrect data substitution. This should be redacted when the database is dumped for testing.

fritzj2:
## Issue Description

Fixes #561 
- The Development team wanted to collect more information from each email to better troubleshoot problems and errors.

## Changes

- Added recipientEmails to emailTracker. This tracks the emails of those who received the email.
- template replaces the subject in the emailTracker. Now the ID of the email template is stored, rather than just the email subject.
- Added body to the emailTracker. This stores the actual body text of the email that was sent to view any errors in the text.
- emailHandler contains recipientEmails, template, and body when adding to the emailTracker object.
- data-scrub.sql has 2 new lines to update any emailtracker objects when the database is dumped. This redacts emails and body text.
- demo_data.py has an EmailTracker object built into the test data for faster testing of the emailTracker changes.

## Testing

New information tracked

- Import test data. Ensure the development environment is being used.
- Use flask run to access the website.
- Go to -> Administration -> Pending forms
- Approve the test forum. This sends a sample email and is tracked in MySQL.
- See further instructions below for accessing MySQL.

Redactions

- Import the test data to get access to the test emailTracker object. Ensure you are in the development environment.
- The data is stored in MySQL, open the MySQL terminal using `mysql -u root -proot`
- If the command above doesn't run, add `--skip-ssl` at the end of it.
- Once in the MySQL database, change the directory to lsf using `USE lsf;` in the MySQL terminal.
- Use `SHOW tables;` to bring up a list of databases that MySQL has access to.
- Use `SELECT * FROM emailtracker;` to bring up the list of data that the emailTracker has recorded so far. Ensure that the test data is present
- After confirming the test data exists, `data_scrub.sql` needs to be changed for terminal support.
- Add `;` every time a line concludes with **END**. The `data_scrub.sql` shouldn't have the `;` when the data is actually being backed up, but won't run in the terminal without them.
- Once the `;` are added, run `source <filepath to data_scrub.sql>;`. This will run the data_scrub.sql file and redact the data.
- Check that the data has been redacted after scrubbing the data. (`SELECT * FROM emailtracker;`)

### DIFF
--- a/app/logic/emailHandler.py
+++ b/app/logic/emailHandler.py
@@ -383,9 +383,8 @@ class emailHandler():
                         date = datetime.today().strftime('%Y-%m-%d'),
                         recipient = recipient,
                         subject = template.subject,
-                        # template = template.body,
                         body = formTemplate,
-                        
+                        recipientEmails = ",".join(message.recipients) 
                         )
         self.send(message)
 
--- a/app/logic/emailHandler.py
+++ b/app/logic/emailHandler.py
@@ -383,8 +383,9 @@ class emailHandler():
                         date = datetime.today().strftime('%Y-%m-%d'),
                         recipient = recipient,
                         subject = template.subject,
+                        templateID = template.emailTemplateID,
+                        recipientEmails = ",".join(message.recipients),
                         body = formTemplate,
-                        recipientEmails = ",".join(message.recipients) 
                         )
         self.send(message)
 
--- a/app/logic/emailHandler.py
+++ b/app/logic/emailHandler.py
@@ -347,11 +347,11 @@ class emailHandler():
 
     # Depending on the parameter 'sendTo', this method will send the email either to the Primary, Secondary, or the Student
     def sendEmail(self, template, sendTo):
+        
         formTemplate = template.body
         formTemplate = self.replaceText(formTemplate)
         if sendTo == "student":
-            message = Message(template.subject,
-                recipients=[self.studentEmail])
+            message = Message(template.subject, recipients = [self.studentEmail])
             recipient = 'Student'
         elif sendTo == "secondary":
             if self.term.isBreak:
@@ -363,8 +363,7 @@ class emailHandler():
                     recipients=supervisorEmails)
                 recipient = 'Secondary Supervisor'
             else:
-                message = Message(template.subject,
-                    recipients=[self.supervisorEmail, self.primaryEmail])
+                message = Message(template.subject, recipients=[self.supervisorEmail, self.primaryEmail])
                 recipient = 'Primary Supervisor'
         elif sendTo == "Labor Office":
             message = Message(template.subject,
@@ -375,18 +374,19 @@ class emailHandler():
                 recipients=[self.supervisorEmail])
             recipient = 'Primary Supervisor'
         elif sendTo == "admin":
-            message = Message(template.subject,
-                recipients=[self.adminEmail])
+            message = Message(template.subject, recipients=[self.adminEmail])
             recipient = 'Admin'
         message.html = formTemplate
-
+        
         newEmailTracker = EmailTracker.create(
                         formID = self.laborStatusForm.laborStatusFormID,
                         date = datetime.today().strftime('%Y-%m-%d'),
                         recipient = recipient,
-                        subject = template.subject
+                        subject = template.subject,
+                        # template = template.body,
+                        body = formTemplate,
+                        
                         )
-
         self.send(message)
 
     # This method is responsible for replacing the keyword form the templates in the database with the data in the laborStatusForm
--- a/app/logic/emailHandler.py
+++ b/app/logic/emailHandler.py
@@ -382,7 +382,6 @@ class emailHandler():
                         formID = self.laborStatusForm.laborStatusFormID,
                         date = datetime.today().strftime('%Y-%m-%d'),
                         recipient = recipient,
-                        subject = template.subject,
                         templateID = template.emailTemplateID,
                         recipientEmails = ",".join(message.recipients),
                         body = formTemplate,
--- a/app/logic/emailHandler.py
+++ b/app/logic/emailHandler.py
@@ -382,7 +382,7 @@ class emailHandler():
                         formID = self.laborStatusForm.laborStatusFormID,
                         date = datetime.today().strftime('%Y-%m-%d'),
                         recipient = recipient,
-                        templateID = template.emailTemplateID,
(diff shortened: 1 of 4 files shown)


app/models/emailTracker.py
```diff
@@ -1,9 +1,14 @@
...
+from app.models.emailTemplate import EmailTemplate
 
 class EmailTracker(baseModel):
     emailTrackerID     = PrimaryKeyField()
-    formID             = ForeignKeyField(LaborStatusForm, on_delete="cascade")               # foreign key to lsf
+    formID             = ForeignKeyField(LaborStatusForm)               # foreign key to lsf
     date               = DateField()
     recipient          = CharField()
-    subject            = CharField()
+    templateID         = ForeignKeyField(EmailTemplate)                 # foreign key to email template
```
MImran2002: just template cause most of the foreignkeys in the models don not have IDs attached to their name only when its a form and even with form foreign keys you don't see this often

app/logic/emailHandler.py
```diff
@@ -375,18 +374,19 @@ def sendEmail(self, template, sendTo):
...
             recipient = 'Admin'
         message.html = formTemplate
-
+        
         newEmailTracker = EmailTracker.create(
                         formID = self.laborStatusForm.laborStatusFormID,
                         date = datetime.today().strftime('%Y-%m-%d'),
                         recipient = recipient,
-                        subject = template.subject
+                        subject = template.subject,
```
MImran2002: isn't subject removed from EmailTracker?

database/demo_data.py
```diff
@@ -609,3 +610,17 @@
...
+# Creates an emailTracker
+#############################
+EmailTracker.insert([{
+            "emailTrackerID": 1000000,
+            "formID": 2,
+            "date": "2026-06-18",
+            "recipient": "test guy",
+            "templateID": 4,
+            "recipientEmails": "coolguy@email.com",
+            "body": "super secret text that you cant read"
```
MImran2002, line 625: A bit more lsf related because still its going to be data use for testing so same with recipient, email too can be berea.edu

MImran2002 (changes requested):

app/logic/emailHandler.py
fritzj2: It should be removed, this has been fixed

app/models/emailTracker.py
fritzj2: Noted. The `templateID` has been changed to `template`.

database/demo_data.py
fritzj2: The `recipient` and `recipientEmails` fields have been altered to better reflect real data.

MImran2002:
As this is a database, back-end issue as well where we are working closely with redaacts and dumping data too. This would need a second reviewer. Go ahead and assign Brian to this too. @fritzj2

MImran2002 (changes requested):
make sure this doesn't fail on the production backup database so you will have insert sql commands in prod-backup.sql I believe for similar database change refer to Sara's CCE Minor production database fix: https://github.com/BCStudentSoftwareDevTeam/celts/pull/1737 and more over don't delete the subject category unless Brian explicitly said so as we don't want the subject in the sever database to lose data as merging that will cause older data to be erase as we remove a data row from it.

fritzj2:
There is now support for prod-backup. The way it does this is by keeping the original fields, but adding the fields at the end which CAN be NULL as the previous data doesn't contain the rows.

app/models/emailTracker.py
```diff
@@ -1,9 +1,15 @@
 from app.models import *
 from app.models.laborStatusForm import LaborStatusForm
+from app.models.emailTemplate import EmailTemplate
 
 class EmailTracker(baseModel):
     emailTrackerID     = PrimaryKeyField()
-    formID             = ForeignKeyField(LaborStatusForm, on_delete="cascade")               # foreign key to lsf
```
MImran2002: can I know why on_delete="cascade' is removed?

database/demo_data.py
```diff
@@ -609,3 +610,17 @@
...
 print(" * laborOfficeNotes added")
+
+
+#############################
+# Creates an emailTracker
+#############################
+EmailTracker.insert([{
+            "emailTrackerID": 1000000,
+            "formID": 2,
+            "date": "2026-06-18",
```
MImran2002: subject needed here

app/models/emailTracker.py
```diff
@@ -1,9 +1,15 @@
...
+    formID             = ForeignKeyField(LaborStatusForm)               # foreign key to lsf
     date               = DateField()
     recipient          = CharField()
     subject            = CharField()
```
MImran2002: because we are moving away from subject being use, we can likely having it as null=true as an indication of subject is empty

MImran2002 (changes requested):

fritzj2:
Currently waiting to ask Brian about a database question:
Do the additional rows in the model need to be null? It depends on how the production data handles the new rows. If it can automatically fill them with NULL or an empty string, then it will be fine; if it won't fill them by itself, then it will need to be explicitly stated as NULL.

Kafui123:
When you try sending an email to your inbox, does it work for you? For some reason, it is not working on my end.

app/models/emailTracker.py
```diff
@@ -1,9 +1,15 @@
 from app.models import *
 from app.models.laborStatusForm import LaborStatusForm
+from app.models.emailTemplate import EmailTemplate
 
 class EmailTracker(baseModel):
     emailTrackerID     = PrimaryKeyField()
-    formID             = ForeignKeyField(LaborStatusForm, on_delete="cascade")               # foreign key to lsf
+    formID             = ForeignKeyField(LaborStatusForm)               # foreign key to lsf
```
Kafui123: I think that it might be a good idea to add a on_delete="cascade" here so in case the labor status form gets deleted the email will get deleted with it as well.

Issue #576, opened 2026-05-20
Facilities Management issue when adjusting labor status form


Karina-Agliullova:
When you try to adjust the labor status form and choose "Facilities Management - Storeroom." It doesn't let you choose any other area in facilities management except "storeroom". Investigate the issue.

<img width="2022" height="1260" alt="Image" src="https://github.com/user-attachments/assets/c2976c01-bcf8-4325-ac35-97c6b9088749" />

Karina-Agliullova:
https://trello.com/c/UEXSvxks/1085-facilities-management-issue-when-adjusting-labor-status-form

Issue #547, opened 2025-10-17
#547 Academic Vice Pres/Provost & Dean Default Error


Meatchema:
Issue:  #547 

Currently, If you create an LSF form that includes SSDT as the department and then try to create an adjustment form for that same form then the department will default to the Academic Vice Pres/Provost & Dean department. We want it so that it defaults to the Computer Science-Student Software Development Team Department instead. This error does not seem to occur with different departments on the original LSF form (ex. choosing Biology Department on the original LSF form)

<img width="1604" height="643" alt="Image" src="https://github.com/user-attachments/assets/0d253df9-d0c9-4eda-95f4-f25e2d6b332c" />

We believe this has to do with the fact that the SSDT Department doesn't have an account or ORG number within MySQL. It was also mentioned that a department number may have been reused that could be causing the Academic Vice Pres/Provost & Dean department to be the default.

PR #593, opened 2026-06-23
Default Error for department 

NYABUTOA:
issue number: #547 Academic Vice Pres/Provost & Dean Default Error #547

Changes:
Updated the adjustment form department default logic to match departments based on both ORG and ACCOUNT, rather than ORG only. This prevents SSDT forms from defaulting to the wrong department when creating an adjustment form.

Testing:
To test this fix, look at what was there in staging and create an LSF form with a department that has a sub-department. After the form is created, open the adjustment form for that same LSF. Check the department dropdown and confirm it defaults to Computer Science-Student Software Development Team instead of Academic Vice Pres/Provost & Dean or the first thing. You can also check this in/home/vscode/lsf/app/controllers/main_routes/alterLSF.py and/home/vscode/lsf/app/templates/main/alterLSF.html. This way, you can see what was changed and added. The biggest thing is that it was only checking for ORG.

### DIFF
--- a/app/controllers/main_routes/alterLSF.py
+++ b/app/controllers/main_routes/alterLSF.py
@@ -46,6 +46,7 @@ def alterLSF(laborStatusKey):
     prefillsupervisorPIDM = form.supervisor.PIDM
     superviser_id = form.supervisor.ID
     prefilldepartment = form.department.ORG
+    prefilldepartmentaccount = form.department.ACCOUNT
     prefillposition = form
     prefilljobtype = form.jobType
     prefillterm = form.termCode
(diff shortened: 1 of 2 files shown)


fritzj2 (changes requested):
Image

The issue with the student software developer is not fixed. Seems to work for other jobs with a similar problem (i.e, Facilities management and Residence Halls.

fritzj2 (approved):
Changes look good.
The reason why the student software developer fails is that it hasn't been implemented on the test data yet. The last backup was before SSDT became its own job, yet TRACY will grab it anyway. This fix means it looks through the database for the position to compare it, but if it cannot find it, it defaults to the first option. This also fixes issue #576

MImran2002:
@johnlolonga19 hey John this issue is tiny fix so not a problem at all but just to be on the side of precaution to not do this unless you are confident it will work.  If you think it will be I would suggest you to ask someone not from your team to merge it after two people have approved not one because this merge doesn't follow proper PR guidelines.

PR #596, opened 2026-06-30
changed migrate_db.sh data to add PositionHistory, and added dummy da…

conwelld:
## Changes

- changed migrate_db.sh data to add PositionHistory, and added demo data to demo_data.py to test data tables to work through mysql

## Testing

- used terminal to access mysql and show tables, and look at position history and changes were made

### DIFF
--- a/app/models/positionHistory.py
+++ b/app/models/positionHistory.py
@@ -7,5 +7,5 @@ class PositionHistory(baseModel):
     status             = CharField()
     WLS                = IntegerField()
     revisiondate       = DateField()
-    Description        = TextField(default="None")
+    Description        = TextField(default=None)
     Department         = ForeignKeyField(Department)               
--- a/app/models/positionHistory.py
+++ b/app/models/positionHistory.py
@@ -8,4 +8,6 @@ class PositionHistory(baseModel):
     revisiondate       = DateField()
     Description        = TextField(default=None)
     Department         = ForeignKeyField(Department)
-PositionHistory._meta.set_primary_key('positioncode_revisiondate_status', CompositeKey('positioncode', 'revisiondate', 'status'))          
+
+class Meta:
+    primary_key = CompositeKey('positioncode', 'revisiondate', 'status')
--- a/app/models/positionHistory.py
+++ b/app/models/positionHistory.py
@@ -2,9 +2,10 @@ from app.models import *
 from app.models.department import Department
 
 class PositionHistory(baseModel):
-    positioncode       = PrimaryKeyField()
+    positionID         = PrimaryKeyField()
+    positioncode       = CharField()
     status             = CharField()
     WLS                = IntegerField()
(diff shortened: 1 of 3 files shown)


app/models/positionHistory.py
```diff
@@ -2,9 +2,10 @@
 from app.models.department import Department
 
 class PositionHistory(baseModel):
-    positioncode       = PrimaryKeyField()
+    positioncode       = CharField()
```
MImran2002, line 5: I look at the conference room diagram and it says positioncode must be the primarykey the inbetween method where you want to add your own primary data would be to have CharField(primarykey=true) like format there is one in the model folder that following the same format.

app/models/positionHistory.py
```diff
@@ -2,9 +2,10 @@
...
     status             = CharField()
     WLS                = IntegerField()
     revisiondate       = DateField()
-    Description        = TextField()
-    Department         = ForeignKeyField(Department)               
+    Description        = TextField(default=None)
+    Department         = ForeignKeyField(Department)
+PositionHistory._meta.set_primary_key('positioncode_revisiondate_status', CompositeKey('positioncode', 'revisiondate', 'status'))
```
MImran2002: Image
using like this is mores standard then doing positionhistoy.meta and so on. This is a tip, if you are implementing something in the model and it is different than usual implementation try looking at other models or even other repos like celts. Your method is another method but not the method SSDT does.

MImran2002: Another was the decision to have them as composite key discussed as I didn't see them on the board. Just want to make sure from my end.

database/demo_data.py
```diff
@@ -609,3 +612,56 @@
...
+        "positioncode": "S61407",
+        "status": "Active",
+        "WLS": 1,
+        "revisiondate": f"{current_year}-07-01",
+        "Description": "",
+        "Department_id": 1
+    },
+    {
+        
+        "positioncode": "S61408",
```
MImran2002, line 838: just cautionary to ensure that to make all sure you are at least having one unique in the composite key.

MImran2002 (changes requested):

app/models/positionHistory.py
```diff
@@ -2,9 +2,12 @@
...
+    positioncode       = CharField()
     status             = CharField()
     WLS                = IntegerField()
     revisiondate       = DateField()
-    Description        = TextField()
-    Department         = ForeignKeyField(Department)               
+    Description        = TextField(default=None)
+    Department         = ForeignKeyField(Department)
+
+class Meta:
```
MImran2002: should be nested within positionhistory

MImran2002 (changes requested):

MImran2002 (changes requested):
resolve the merge conflict

MImran2002 (approved):

PR #598, opened 2026-07-01
Adding Position History & Dummy Data for Department Members 

ArtemKurasov:
# Changes 

- Added the position history file to the database (can be found in the `app/models/positionHistory.py` file). 
- Added some dummy data for Department Members to the `demo_data.py` file.

# Testing 

- Test if the dummy data is in the database: 
   - type `mysql -u root -proot --skip-ssl` to open MySQL
   - then enter `USE lsf;` to go to the LSF database specifically
   - run "SHOW TABLES;" to view all the entities that exist in the LSF database
   - type `SELECT * FROM supervisordepartment;` to select the table representing Department Members 
   - if there's something similar to the screenshot below, everything works properly!
   - enter "EXIT;" to exit MySQL
   - run Flask to ensure that the dummy data does not break the website

image

### DIFF
--- a/app/config/contributors.yaml
+++ b/app/config/contributors.yaml
@@ -59,4 +59,7 @@ contributors:
   - name: "Oluwagbayi Makinde"
     username: "makindeo"
     year: 2024
+  - name: "Bhushan Sah "
+    username: "sahb"
+    year: 2029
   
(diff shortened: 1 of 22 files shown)


MImran2002:
Since this branch pulled in both development and PeeweeAttribute, it can create messy PR history so the peeweeattribute has overwritten the local changes as a result this PR will be close

PR #637, opened 2026-07-08
Members only

BhushanSah:
Made the Improvement for the Members Card
Fixes #603


fritzj2 (changes requested):
- This branch should not be merged into development, it should be merged into department-portal-base
- The HTML file needs to use the variable `labor_coordinator`, not `labor_coordinators`. this is why none are appearing
- Try to use try and except in the event that it cannot find something correctly to prevent the page from just crashing.
- Add a PR description so I know what to look for

PR #636, opened 2026-07-07
Wire Current Allocation card to real backend data (#602)

DanielRukwasha:
Related to #602

Replaces the hardcoded placeholder numbers on the Department Portal's 'Current Allocation' card with real data:
- Fixed \database/migrate_db.sh\, which was missing \pp.models.allocation.Allocation\ entirely — the \llocation\ table was never being created via migrations.
- Added demo \Allocation\ seed rows per department (none existed before).
- \departmentPortal()\ now queries the department's \Allocation\ row for the current open term, and computes real 'used' counts per hour-band (10/12/15/20 primary, 5/10 secondary) by counting non-denied \LaborStatusForm\s — reusing the same denied-filter pattern already used elsewhere in the app (\statusFormFunctions.py\).
- Added a best-effort 'break hours used' calculation (summed from \contractHours\ on break-term forms), since nothing previously tracked this.
- Added 'Primary' / 'Secondary' section titles above the hour-band breakdown lists for clarity.
- Removed the non-functional 'Request Allocation' button (no destination page exists yet) and relabeled 'AY' to 'Term' in the term display.

Not yet done / out of scope for this PR: the 'Manage Allocations' button still points nowhere (destination page depends on other open issues), and there's no access-control check restricting a department's allocation data to Labor admins + that department's own staff.

Tested locally end-to-end in the devcontainer against real seeded data across multiple departments.

PR #638, opened 2026-07-08
Display allocation warning during pending LSF approval (#622)

DanielRukwasha:
Closes #622

Adds a non-blocking over-allocation warning to the pending-LSF approval confirmation modal, so Labor Office admins can see the allocation impact before approving.

- Extracted the allocation-counting logic from #602's Department Portal card into a shared \pp/logic/allocation.py\ (\getAllocationSummary\, \getAllocationWarning\), so this reuses the exact same 'used' counting (non-denied LaborStatusForms) instead of duplicating queries.
- \modal_approval_and_denial_data()\ now returns one allocation warning per unique department among the selected forms, alongside the existing per-form table data.
- The approval modal (\pendingApprovalModal.html\ / \llPendingForms.js\) renders a warning box per department: blue if within allocation, amber if any category is over. Positions and Break Hours are highlighted **independently** per the issue's requirement ('highlight any allocation categories that exceed'), since a department can be over on one and fine on the other.
- Fixed a pre-existing demo-data gap found while testing: no student/supervisor was ever marked \isActive\, so the approval checkbox (which requires both to be active) never rendered at all in this environment.

Branched off \llocation-card-real-data-ruk\ (not \llocation_card\ directly) since this depends on the allocation logic from #636, which hasn't merged yet.

Tested via direct API calls and in-browser: confirmed the warning shows correct numbers, confirmed the red/amber highlighting triggers independently for positions-only and break-hours-only over-allocation scenarios, and confirmed the full approve flow completes correctly end-to-end.

### DIFF
--- a/app/controllers/main_routes/departmentPortal.py
+++ b/app/controllers/main_routes/departmentPortal.py
@@ -1,0 +1,1 @@
+from flask import render_template
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -1,16 +1,12 @@
 from flask import render_template, request, json, redirect, url_for, send_file, g, flash, jsonify
-from peewee import JOIN, DoesNotExist, fn
+from peewee import JOIN, DoesNotExist
 from functools import reduce
 import operator
-from app.models.allocation import Allocation
 from app.models.department import Department
 from app.models.supervisor import Supervisor
 from app.models.supervisorDepartment import SupervisorDepartment
 from app.models.student import Student
-from app.models.laborStatusForm import LaborStatusForm
 from app.models.formHistory import FormHistory
-from app.models.term import Term
-from app.models.allocation import Allocation
 from app.controllers.admin_routes.allPendingForms import checkAdjustment
 from app.controllers.main_routes import main_bp
 from app.logic.download import CSVMaker, saveFormSearchResult, retrieveFormSearchResult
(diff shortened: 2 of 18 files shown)


MImran2002:
The warning is not popping up when i follow the steps to create that warning.

MImran2002 (changes requested):
Warning not showing

bakobagassas (commented):
Image 


This part of the warning is a little bit confusing.  **" Positions: 7 / 20 allocated (14 remaining) ⚠ Over allocation" **  It says we have 13 positions remaining but then it also says that we are over allocating. I understand that we indeed have only two 10 hours positions for the computer science department but then we have up 20 positions in total available accross all hours (including the two 10 hours positions)

I think it would be less confusing if we actually let them know that the warning is only for their 10 hours positions and then after that we display that they are still fine for the rest (like 15hrs, etc) and that they have 7/20

RueHaile:
> Image
> This part of the warning is a little bit confusing. **" Positions: 7 / 20 allocated (14 remaining) ⚠ Over allocation" ** It says we have 13 positions remaining but then it also says that we are over allocating. I understand that we indeed have only two 10 hours positions for the computer science department but then we have up 20 positions in total available accross all hours (including the two 10 hours positions)
> 
> I think it would be less confusing if we actually let them know that the warning is only for their 10 hours positions and then after that we display that they are still fine for the rest (like 15hrs, etc) and that they have 7/20

done
image

PR #639, opened 2026-07-08
Add Positions Card into the Department Portal

fritzj2:
## Issue Description

Fixes issue #604 
- Add a card to a position on the manage departments page

## Additions

- Screenshot 2026-07-10 114154

- Added 3 Cards to the manage department page
- The position card has a list of `positionHistory` Positions and their WLS level
- The list of positions takes you to a link where the specific position info should be (but currently isn't implemented)
- The "View all positions" button takes you to where the " Manage/View Positions page will lead.
- Failsafe for no positions in a department ("No positions for this department" appears instead)
- Failsafe for an overload of positions (and x amount more positions)
- `departmentPortal.css` handles all of the styling for the cards. 

## Changes

- Department portal Department selector dropdown no longer takes up the entire screen
- Cards have a variable size that scales with the screen
- The cards stack on top of each other when the screen size is too small.
- The cards do _not_ appear if a department has not been selected.
- `migrate_db.sh` contains an extra line to accommodate the positionHistory being added to the table.
- `positionHistory.py` contains 1 extra line that now contains the `positionTitle`
- Additional files can be found from a development PR merge, which has not yet been implemented into department-portal-base.

## Testing

- Reset the database to include the changes to the `demo_data.py`.
- Booting into the portal using flask run allows you to see if the card is there.
- 10 positions should load for the CS department ( 7 + 3 more), while 1 should load for the Mathematics department
  - The mathematics department has 2 positions assigned to it; however, only 1 is active, so only one loads.
- The demo_data.py file has been updated to include `positionHistory` data within it.
  - The position info page does not yet exist and will return a 404 error when selected; however, the URL still does exist.
- The "View All Positions" button will give you a 404 error when selected, as the webpage for it doesn't exist yet; the URL still exists.
- Mess with the screen size to ensure that the cards do not have a static size. The 2 extra cards will go side by side when the screen is small enough.
- Ensure that the largest size of the department doesn't extend out of the dropdown bar
- This data will not load anything on staging data currently, as nothing has been allocated to positionHistory yet.

### DIFF
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -16,7 +16,6 @@ from app.logic.search import getDepartmentsForSupervisor, searchPerson, searchSu
 from app.login_manager import require_login, logout
 from app.logic.getTableData import getDatatableData
 from app.logic.banner import Banner
-from app.logic.tracy import Tracy
 from app.models.positionHistory import PositionHistory
 
 
@@ -54,12 +53,9 @@ def supervisorPortal():
 @main_bp.route('/department/<org>', methods=['GET'])
 @main_bp.route('/department/<org>/<account>', methods=['GET'])
 def departmentPortal(org=None,account=None):
-    if org and account:
-        try:
-            dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
-        except (NameError, DoesNotExist):
-            dept = None
-    else:
+    try:
+        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+    except (NameError, DoesNotExist):
         dept = None
 
 
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -78,8 +78,8 @@ def departmentPortal(org=None,account=None):
         for i in pos:
             positions.append(i.POSN_TITLE + ": " + "(WLS " + i.WLS + ")")
             try:
-                pos_his_obj = PositionHistory.get(PositionHistory.positioncode == i.POSN_CODE)
-                pos_his.append(str(pos_his_obj.positioncode) + str(pos_his_obj.revisiondate))
+                pos_his_obj = PositionHistory.get(PositionHistory.positionCode == i.POSN_CODE)
+                pos_his.append(str(pos_his_obj.positionCode) + str(pos_his_obj.revisionDate))
             except:
                 pos_his.append("#")
 
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -69,10 +69,10 @@ def departmentPortal(org=None,account=None):
         departments = list(getDepartmentsForSupervisor(g.currentUser).order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
 
         
-    positions = list(PositionHistory.select().where(PositionHistory.department == dept).order_by(PositionHistory.positionTitle.asc())) if dept else []
+    positions = list(PositionHistory.select().where(PositionHistory.department == dept, PositionHistory.status == "Active").order_by(PositionHistory.positionTitle.asc())) if dept else []
     positionsList = []
     posUrl = []
-    if positions == []:
+    if not positions:
         positionsList = ["No Positions for this Department"]
     else:
         for i in positions:
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -16,6 +16,9 @@ from app.logic.search import getDepartmentsForSupervisor, searchPerson, searchSu
 from app.login_manager import require_login, logout
 from app.logic.getTableData import getDatatableData
 from app.logic.banner import Banner
+from app.logic.tracy import Tracy
+from app.models.positionHistory import PositionHistory
+
 
 @main_bp.route('/logout', methods=['GET'])
 def triggerLogout():
@@ -51,11 +54,13 @@ def supervisorPortal():
 @main_bp.route('/department/<org>', methods=['GET'])
 @main_bp.route('/department/<org>/<account>', methods=['GET'])
 def departmentPortal(org=None,account=None):
-    try:
-        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
-    except (NameError, DoesNotExist):
+    if org and account:
+        try:
+            dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+        except (NameError, DoesNotExist):
+            dept = None
+    else:
         dept = None
-
 
 
     if g.currentUser.isLaborAdmin:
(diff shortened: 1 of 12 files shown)


conwelld:
Looks Good

app/controllers/main_routes/main_routes.py
```diff
@@ -51,21 +54,40 @@ def supervisorPortal():
...
 
 
-
     if g.currentUser.isLaborAdmin:
         departments = list(Department.select().order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
     else:
         departments = list(getDepartmentsForSupervisor(g.currentUser).order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
 
+        
+    pos = Tracy().getPositionsFromDepartment(org, account)
```
MImran2002: this is going to work from test data I believe and using tracy means we are setting the environment to staging or production and that means the demo data is void as using tracy means connecting to tracy which means we are using production data not demo data.

app/controllers/main_routes/main_routes.py
```diff
@@ -51,21 +54,40 @@ def supervisorPortal():
 @main_bp.route('/department/<org>', methods=['GET'])
 @main_bp.route('/department/<org>/<account>', methods=['GET'])
 def departmentPortal(org=None,account=None):
-    try:
-        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
-    except (NameError, DoesNotExist):
+    if org and account:
```
MImran2002: is there a rationale for this as the route /department/<org> will return only org so that will break the conditional. which means there are no deparment

app/controllers/main_routes/main_routes.py
```diff
@@ -51,21 +54,40 @@ def supervisorPortal():
...
-
     if g.currentUser.isLaborAdmin:
         departments = list(Department.select().order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
     else:
         departments = list(getDepartmentsForSupervisor(g.currentUser).order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
 
+        
+    pos = Tracy().getPositionsFromDepartment(org, account)
+    positions = []
+    pos_his = []
```
MImran2002: coding convention we use camel

app/controllers/main_routes/main_routes.py
```diff
@@ -51,21 +54,40 @@ def supervisorPortal():
...
+    if pos == []:
```
MImran2002: not a descriptive variable

MImran2002: `if not pos` is more readable

app/controllers/main_routes/main_routes.py
```diff
@@ -51,21 +54,40 @@ def supervisorPortal():
...
+        positions = ["No Positions for this Department"]
+    else:
+        for i in pos:
```
MImran2002: not a descriptive i

app/controllers/main_routes/main_routes.py
```diff
@@ -51,21 +54,40 @@ def supervisorPortal():
...
+            positions.append(i.POSN_TITLE + ": " + "(WLS " + i.WLS + ")")
+            try:
+                pos_his_obj = PositionHistory.get(PositionHistory.positioncode == i.POSN_CODE)
```
MImran2002: here too not convention and not descriptive

app/controllers/main_routes/main_routes.py
```diff
@@ -51,21 +54,40 @@ def supervisorPortal():
...
+                pos_his.append(str(pos_his_obj.positioncode) + str(pos_his_obj.revisiondate))
+            except:
+                pos_his.append("#")
+
     return render_template('main/departmentPortal.html', 
                            departments = departments,
-                           department = dept)
+                           department = dept,
+                           positions = positions,
+                           pos_his = pos_his)
```
MImran2002: here too

app/templates/main/departmentPortal.html
```diff
@@ -9,7 +9,7 @@
 <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% else %} Choose a Department: {% endif %} </h2>
 
   <!--selectpicker for department-->
-  <div class="form-group">
+  <div class="form-group" style="width:65rem; margin-left: auto; margin-right:auto">
```
MImran2002: use bootstrap class for breakpoints 

Bootstrap Version | Left Alignment | Right Alignment
Bootstrap 5 | float-start | float-end
Bootstrap 4 | float-left | float-right
Bootstrap 3 | pull-left | pull-right

base on what bootstrap you chose. Those might be wrong too I would suggest you to research into this to suit the bootstrap version we use.

app/templates/main/departmentPortal.html
```diff
@@ -24,4 +24,53 @@ <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% e
     </select>
 
   </div>
+
+  {% if department %}
+<div class="card-group" style="display:inline-flex; gap: 1rem">
```
MImran2002: bootstrap

app/templates/main/departmentPortal.html
```diff
@@ -24,4 +24,53 @@ <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% e
...
+     
+  <section class="card "style="width:90%; border-radius: 1rem;" aria-labelledby="current_allocation-title">
```
MImran2002: all your style should be bootstrap unless it is really needed even then we only use inline styling when there is no css file for it.

database/demo_data.py
```diff
@@ -609,3 +610,63 @@
...
 print(" * laborOfficeNotes added")
+
+#############################
+# Position History
+#############################
+
+positionhistory = [
+    {
+        "positioncode": "S61407",
+        "status": "Active",
```
MImran2002, line 757: beware of current year Brian have told us about this. I think he has said this is affecting the test suite look into the recent change brian did and see if its use or been replaced.

MImran2002 (changes requested):

app/templates/main/departmentPortal.html
```diff
@@ -24,4 +25,49 @@ <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% e
...
+        <div class="media-body media-middle">
+          <h1>Positions</h1>
+        </div>
+
+      </div>
+      <ul style="line-height:2;">
+      {% for p in positions %}
+        {% if 7 > loop.index0 %}
+          <li class="card-text">
+            <a href= "/department/{{ department.ORG }}/{{ department.ACCOUNT }}/managepositions/{{posUrl[loop.index0]}}">{{ p }}</a> </li>
```
BetterRocket: Should probably not make the "No Positions for this Department" text a hyperlink .

Image

BetterRocket:
Also add enough positions in demo_data.py to see the the "and _ more..." text on the card

app/templates/main/departmentPortal.html
```diff
@@ -24,4 +25,72 @@ <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% e
...
+                    <li class="card-text">
+                    <p> {{ p }} </p>
+                    </li>
+                  {% else %}
+                    <li class="card-text">
+                      <a href= "/department/{{ department.ORG }}/{{ department.ACCOUNT }}/positions/{{posUrl[loop.index0]}}">{{ p }}</a>
+                    </li>
+                  {% endif %}
+
+              {% elif 8 == loop.index0 %}
```
MImran2002: Bug here you first if condition ends at index 6 (0-6) then if there is 7 the seven is not counted and only will start considering at 8.

MImran2002 (changes requested):
see the request

app/controllers/main_routes/main_routes.py
```diff
@@ -57,15 +59,28 @@ def departmentPortal(org=None,account=None):
...
         departments = list(Department.select().order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
     else:
         departments = list(getDepartmentsForSupervisor(g.currentUser).order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
 
+        
+    positions = list(PositionHistory.select().where(PositionHistory.department == dept, PositionHistory.status == "Active").order_by(PositionHistory.positionTitle.asc())) if dept else []
+    positionsList = []
+    posUrl = []
+    if not positions:
+        positionsList = ["No Positions for this Department"]
```
MImran2002: No active positions in this department

MImran2002 (changes requested):

MImran2002 (approved):
Approved:
CSS file created because LSF bootstrap is outdated for some bootstrap utility classes
Peewee logic is correct
positionTitle attribute created in reference to what was discussed and sketched out in the conference room

These are some that I would change personally as good coding practice:
-posUrl = [] should have been posURL
-not recommended practice for designating No active positions in this department in the backend comparing it in the frontend then using it when it can be an empty string and use as a template as using No active positions in this department in the backend and conditional checking it as No active positions in this department can create bug.

PR #642, opened 2026-07-08
Adding the Manage Members Page + Fixing the Demo Data Discrepancies 

ArtemKurasov:
We have added the _Manage Members_ page, which can be accessed through the _Members_ card on the _Department Portal._ Besides that, we also eliminated demo data issues, which were breaking the database. 

**WHAT'S WORKING:** 
- The names of supervisors are correctly pulled from the demo data. 
- Supervisees are properly displayed (on the table, `active` means the student has approved their labor status form, and `pending` means they have not done that yet).
- The Assign Coordinator switch works! 
- The Ban/Unban Button also works! 
- You can remove members by clicking on the trash icon to the right (YOU CAN'T ADD THEM BACK YET)
- The Add Members dropdown/button _technically_ works.
- The Ban/Unban button and the Remove Member button are displayed ONLY FOR ADMINS (as it is supposed to be).

**WHAT'S _NOT_ WORKING:**
- You can only add members who have not been initially part of the chosen department (for example, Megan Hoffman is the only person that can be added to the Computer Science Department, because she is not part of the CS department in the database).


MImran2002 (approved):
approved with conditions

PR #643, opened 2026-07-09
Demo data fix dpb

conwelld:
## Issue Description

Fixes issue regarding demo data not following Brian's variable naming conventions, as well as fixes the semantic errors within allocations as well as added demo data to department-portal-base

- department-portal-base had no demo data to match the new models in Brian's branch, I added the dummy data from PeeWeeAttribute and fixed the naming conventions to match Brian's I also fixed a database error within the data where allocations called a nonexistent term within Term.termcode. I also fixed the naming of the department members data because its naming was unclear as well as position history 

## Testing

- I reset database with no issues and was able to flask run with no issues

### DIFF
--- a/database/demo_data.py
+++ b/database/demo_data.py
@@ -662,7 +662,6 @@ supervisorDepartmentMembers = [
 SupervisorDepartment.insert_many(supervisorDepartmentMembers).on_conflict_replace().execute()
 print(" * Department members added")
 print(f"termCode_id being used: {202500!r}")
-print(list(Term.select(Term.termCode)))
 
 ############################
 # Allocation Dummy Data:
@@ -747,8 +746,7 @@ allocations = [
     ]
 Allocation.insert_many(allocations).on_conflict_replace().execute()
 
-print("Data insertion complete :)")
-print(PositionHistory())
+print(" * allocation added")
 
 
 #############################
@@ -801,3 +799,4 @@ positionHistory = [
     
 ]
 PositionHistory.insert_many(positionHistory).on_conflict_replace().execute()
+print(" * position history added")
--- a/database/demo_data.py
+++ b/database/demo_data.py
@@ -2,7 +2,6 @@
 Chech phpmyadmin to see if your changes are reflected
 This file will need to be changed if the format of models changes (new fields, dropping fields, renaming...)'''
 
-from datetime import *
 from app import app
 
 from app.models.Tracy import db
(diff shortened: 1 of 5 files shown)


database/demo_data.py
```diff
@@ -618,3 +621,183 @@
...
+    "primary_20":       4,
+    "secondary_5":      5,
+    "secondary_10":     1,
+    "breakHours":       900,
+    },
+                
+    ]
+Allocation.insert_many(allocations).on_conflict_replace().execute()
+
+print("Data insertion complete :)")
```
MImran2002: print(" * allocation added") should be the correct print out not with a smiling face :)

database/demo_data.py
```diff
@@ -618,3 +621,183 @@
...
+        "positionCode": "S61410",
+        "status": "Inactive",
+        "wls":2,
+        "revisionDate" : f"2025-01-01",
+        "description": "",
+        "department" : 1
+        },
+    
+]
+PositionHistory.insert_many(positionHistory).on_conflict_replace().execute()
```
MImran2002: after this add print(" * position history added") just to be consistent

database/demo_data.py
```diff
@@ -618,3 +621,183 @@
...
+    "secondary_5":      5,
+    "secondary_10":     1,
+    "breakHours":       900,
+    },
+                
+    ]
+Allocation.insert_many(allocations).on_conflict_replace().execute()
+
+print("Data insertion complete :)")
+print(PositionHistory())
```
MImran2002: I don't think this is needed

database/demo_data.py
```diff
@@ -618,3 +621,183 @@
...
+        "supervisor": "B00841417",
+        "department": 1,
+        "isCoordinator": True
+    }
+]
+
+SupervisorDepartment.insert_many(supervisorDepartmentMembers).on_conflict_replace().execute()
+print(" * Department members added")
+print(f"termCode_id being used: {202500!r}")
+print(list(Term.select(Term.termCode)))
```
MImran2002: not needed print statement

MImran2002 (changes requested):

conwelld:
I believe it should be fixed now

database/demo_data.py
MImran2002: as this is no longer use anymore we can remove this and today

MImran2002: this as well both today and current_year

MImran2002: as we have deleted today and current_year this module is no longer needed too so we delete this.

MImran2002 (changes requested):

MImran2002 (approved):
looks good test all passed and the demo data looks good

PR #646, opened 2026-07-13
Add over-allocation warning on labor status form (#615)

DanielRukwasha:
Closes #615. Shows a non-blocking allocation warning when creating a labor status form that would put the department over its allocation for the selected job type/hours band.


app/controllers/main_routes/laborStatusForm.py
```diff
@@ -170,6 +171,15 @@ def checkTotalHours(termCode, student, hours):
     totalHours = totalHours + int(hours)
     return json.dumps(totalHours)
 
+@main_bp.route("/laborstatusform/checkallocation/<departmentOrg>/<departmentAcct>/<jobType>/<hours>", methods=["GET"])
```
MImran2002: this should be better as you are sending four data through the url instead from the js function you can pack those data as a variable data dictionary and send it over to here you can unpack it with request.arg i believe.

app/controllers/main_routes/laborStatusForm.py
```diff
@@ -170,6 +171,15 @@ def checkTotalHours(termCode, student, hours):
...
+def checkAllocation(departmentOrg, departmentAcct, jobType, hours):
+    """ Checks the department's allocation status for the hour-band being submitted. """
+    dept = Department.get_or_none(Department.ORG == departmentOrg, Department.ACCOUNT == departmentAcct)
+    if not dept:
+        return jsonify(None)
```
MImran2002: this should return an error so that when ajax pick up the error it can say why or by using redirect to tell the suer their access is denied.

app/controllers/main_routes/main_routes.py
MImran2002, line 60: all the changes underneath the departmentportal is not under your issue scope delete them.

app/logic/allocation.py
```diff
@@ -0,0 +1,123 @@
...
+    ('primary_15', 'Primary', 15),
+    ('primary_20', 'Primary', 20),
+    ('secondary_5', 'Secondary', 5),
+    ('secondary_10', 'Secondary', 10),
+]
+
+BAND_LABELS = {fieldName: f"{hours} Hour {jobType}" for fieldName, jobType, hours in ALLOCATION_BAND_FIELDS}
+
+
+def getAllocationSummary(dept, term):
```
MImran2002: in this folder the only 2 function you need is the allocation function that goes and grab the allocation utilize and break hours which you should reach out to julius and ask the functions and just copy and paste and you can use that where else finding what already exist you can get it from allocationhisotry model.

app/logic/allPendingForms.py
```diff
@@ -80,6 +84,16 @@ def saveStatus(new_status, formHistoryIds, currentUser):
         print("Error preparing form for status update:", e)
         return jsonify({"success": False}), 500
 
+    # After approving, let the admin know right away if any affected department
+    # is now over its allocated positions or break hours (informational only).
+    for dept in approvedDepartments.values():
+        warning = getAllocationWarning(dept, g.openTerm)
```
MImran2002, line 90: you have two warning you are using a flash message and a display in the model of approving remove the flash message part as we are moving away from it instead choose the modal warning code that you have done.

app/static/js/allPendingForms.js
```diff
@@ -112,6 +112,40 @@ function updateApproveTableData(returned_details) {
...
+    var bandDetail = '';
+    if (w.overAllocatedBands && w.overAllocatedBands.length > 0) {
+      var bandStrings = w.overAllocatedBands.map(function(b) {
+        return b.label + ' (' + b.used + ' used / ' + b.allocated + ' allocated)';
+      });
+      bandDetail = '<br><span style="' + overStyle + '">Over on: ' + bandStrings.join(', ') + '</span>';
+    }
+    var html = '<div class="alert ' + boxClass + '" role="alert">' +
+      '<strong>' + w.departmentName + ' Allocation</strong><br>' +
+      '<span style="' + positionsStyle + '">Total Positions (all bands): ' + w.totalPositionsUsed + ' / ' + w.totalPositionsAllocated +
```
MImran2002: Image
normal text should be black not dark yellow and you show the allocation and a exclaimation with overallocation remove that. we should only show that warning when it is overallocated so instead of an exclaimation mark and over allocation turn the title into ,department> Over Allocation Warning

app/static/js/laborStatusForm.js
```diff
@@ -336,6 +337,40 @@ function checkCompliance(obj) {
...
+  var hours = $("#selectedHoursPerWeek").val();
+
+  if (!departmentOrg || !jobType || !hours) {
+    return;
+  }
+
+  var url = "/laborstatusform/checkallocation/" + departmentOrg + "/" + departmentAcct + "/" + jobType + "/" + hours;
+  $.ajax({
+    url: url,
+    dataType: "json",
```
MImran2002, line 368: bundle the dapartmentOrg, departmentAcct, jobtype and hours as a data dictionary instead of adding it into the url

app/templates/main/departmentPortal.html
```diff
@@ -26,29 +26,94 @@ <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% e
 
   </div>
 
-  {% if department %}
+{% if department %}
```
MImran2002: you dont need to have deopartment portal html you are just focusing on the warning.

MImran2002 (changes requested):

app/controllers/main_routes/laborStatusForm.py
```diff
@@ -14,14 +14,15 @@
 from flask import json, jsonify
 from flask import request
 from datetime import datetime, date, timedelta
-from flask import Flask, redirect, url_for, flash
+from flask import Flask, redirect, url_for, flash, g
 from app.logic.emailHandler import*
```
MImran2002, line 18: we don't use the terminology bands

app/controllers/main_routes/laborStatusForm.py
```diff
@@ -170,6 +171,41 @@ def checkTotalHours(termCode, student, hours):
     totalHours = totalHours + int(hours)
     return json.dumps(totalHours)
 
+@main_bp.route("/laborstatusform/checkallocation", methods=["GET"])
```
MImran2002, line 174: UI test: nothing is showing up in the table: 

Image

app/controllers/main_routes/laborStatusForm.py
```diff
@@ -170,6 +171,41 @@ def checkTotalHours(termCode, student, hours):
...
+def checkAllocation():
+    """ Checks the department's allocation status for the hour-band being submitted. """
+    departmentOrg = request.args.get("departmentOrg")
+    departmentAcct = request.args.get("departmentAcct")
+    jobType = request.args.get("jobType")
+    hours = request.args.get("hours")
```
MImran2002, line 180: Ui Test: even if theres over allocated no warning show up for individual department or multiple department: 

Image

MImran2002 (changes requested):

PR #649, opened 2026-07-20
Adding the Manage Members Page

brightfietsop-ux:
We have added the Manage Members page, which can be accessed through the Members card on the Department Portal.

**WHAT'S WORKING:**

- The names of supervisors are correctly pulled from the demo data.
- Supervisees are properly displayed (on the table, active means the student has approved their labor status form, and - pending means they have not done that yet).
- The Assign Coordinator switch works!
- The Ban/Unban Button also works!
- You can remove members by clicking on the trash icon to the right 
- The Add Members dropdown/button technically works.
- The Ban/Unban button and the Remove Member button are displayed ONLY FOR ADMINS (as it is supposed to be).


app/templates/main/managepositions.html
```diff
@@ -0,0 +1,90 @@
+{% extends "base.html" %}
```
MImran2002: this page shouldn't even be here outside your issue scope

app/templates/main/supervisorPortal.html
```diff
@@ -227,7 +227,7 @@ <h1 class="fs-1" data-toggle="collapse">Form Search</h1>
 </div>
 <br>
 <div style="position: relative; z-index: 1;" id="sortOptions">
-  <div class="container-fluid" style="margin-bottom: -40px; float: right;">
+  <div class="container-fluid" style="margin-bottom: 15px; float: right;">
```
MImran2002: we revert this as this is beyond the scope.

database/reset_database.sh.bak
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
MImran2002: we don't create new scripts for the database, this is outside of your scope

database/reset_database.sh
```diff
@@ -8,7 +8,7 @@ if [ "$1" != "test" ] && [ "$1" != "from-backup" ]; then
 fi
 
 cd database;
-
+
```
MImran2002: we can remove those spaces.

db_test.py
```diff
@@ -14,8 +14,8 @@
 pyodbc_uri = 'DRIVER=FreeTDS;SERVER={};PORT=1433;DATABASE={};UID={};PWD={};TDS_Version=7.4;'.format(details['server'],details['db'],details['user'],details['password'])
```
MImran2002, line 14: this change seems to be out of your issue scope too.

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -1,31 +0,0 @@
-function addSupervisorToDepartment(supervisorID, departmentID, callback=() => {}) {
```
MImran2002, line 1: delete the file if you have removed all the code

app/static/css/managepositions.css
```diff
@@ -0,0 +1,15 @@
+.width-12{
+    width:12%;
+}
+
```
MImran2002: these codes are outside of your issue scope

app/models/supervisor.py
```diff
@@ -13,10 +13,11 @@ class Supervisor(baseModel):
...
 
-    legal_name      = CharField(null=True)
-    preferred_name  = CharField(null=True)
-    isActive        = BooleanField(default=False)
-    isBanned        = BooleanField(default=False)
+    legal_name          = CharField(null=True)
+    preferred_name      = CharField(null=True)
+    isActive            = BooleanField(default=False)
+    isBanned            = BooleanField(default=False)
+
```
MImran2002, line 20: undo what has been changed here

app/logic/tracy.py
```diff
@@ -142,6 +142,14 @@ def getStudentsFromBNumberSearch(self, bnum_part: str):
         students = STUDATA.query.filter(STUDATA.ID.like(bnum_search)).all()
         return students
 
+    def getSupervisorsFromBNumberSearch(self, bnum_part: str):
```
MImran2002: no longer needed

MImran2002 (changes requested):

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+from app.controllers.main_routes import main_bp
+from app.logic.download import CSVMaker, saveFormSearchResult, retrieveFormSearchResult
+from app.logic.search import getDepartmentsForSupervisor, searchPerson, searchSupervisorPortal
+from app.login_manager import require_login, logout
+from app.logic.getTableData import getDatatableData
+from app.logic.banner import Banner
+from flask import abort
+from app.logic.manageMembers import supervisorsDbToDict, currentAcademicYear
+from app.logic.search import limitSearchByUserDepartment, studentDbToDict, usernameFromEmail
+from app.logic.manageMembers import getCurrentDepartment,getDepartmentMembers,getReleasedFormIds,getStudentCounts,attachPositionCounts
```
BrianRamsay: Let's sort our imports. Generally, system python, then application libraries, then our code. And for our code, keep it sorted with models, logic, etc grouped together.

e.g.
```
from functools import reduce
import operator

from flask import abort
from peewee import stuff

from app.models.student import student

File: app/controllers/main_routes/departmentPortal.py [hunk 11]
```diff
@@ -1 +1,172 @@
...
+
+@main_bp.route('/department/<org>/<account>/members', methods=['GET'])
+def manageMembers(org=None, account=None):
+    """Generates the Manage Members page."""
+    currentUser = require_login()
+    if not currentUser or not currentUser.supervisor:
+        if currentUser.student:
```
BrianRamsay: This would break if currentUser is empty, which is technically possible,  but won't happen in our app. Let's use g.currentUser, which will always exist and will let us simplify this. What are the cases where we would actually get a 403 here? You are either a supervisor or student, in our system, so if `not currentUser.supervisor` is true, so is `currentUser.student`.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+            return redirect(url_for('main.laborhistory', id=currentUser.student.ID))
+        return render_template('errors/403.html'), 403
+
+    currentSupervisor = Supervisor.get(Supervisor.ID == currentUser.supervisor)
```
BrianRamsay: This is redundant, as `currentUser.superviser` is already the same Supervisor object you have after the query.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+
+    members = getDepartmentMembers(dept)
+    counts = getStudentCounts(dept)
+    members = attachPositionCounts(members, counts)
+
+    return render_template(
+        'main/manageMembers.html',
+        members=members,
+        department=dept,
+        currentSupervisor=currentSupervisor,
```
BrianRamsay: Don't need to send this, as the template can already access g.currentUser, which has supervisor

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+        currentAcademicYear=currentAcademicYear()
+    )
+
+
+
+@main_bp.route('/members/search/<query>',  methods=['GET'])
+def searchMember(query=None):
```
BrianRamsay, line 98: There are already several places that search students or supervisors. We should try and use whatever those are using. If the logic needs to change, it can probably change there as well.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+    supervisors = sorted(supervisors, key=lambda f: f['firstName'] + f['lastName'])
+
+    return jsonify(supervisors)
+
+@main_bp.route('/members/coordinator_switch', methods=['POST'])
+def coordinatorSwitch():
+    """
+    Assigns or unassignes a supervisor as a Labor Coordinator. 
+    """
+    data = request.get_json()
```
BrianRamsay: we don't need to convert the request to json in order to use it. `request.form.get` is the standard method

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+    supervisorID = data.get("supervisorID")
+    isCoordinator = data.get("isCoordinator")
+
+    departmentID = session.get('current_department_id')
```
BrianRamsay: Using the session to store our department id can cause confusing behavior when using multiple tabs. I would prefer to pass the department id with the request

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+        recordedSupervisors = list(map(supervisorsDbToDict, results))
+        
+
+    # combine lists, remove duplicates, and then sort
+    supervisors = list({v['bnumber']:v for v in (recordedSupervisors)}.values())
+    supervisors = sorted(supervisors, key=lambda f: f['firstName'] + f['lastName'])
+
+    return jsonify(supervisors)
+
+@main_bp.route('/members/coordinator_switch', methods=['POST'])
```
BrianRamsay: Let's put the verb first. `switch_coordinator`, or probably more accurately `update_coordinator`

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+        (SupervisorDepartment.department == departmentID)
+    )
+    member.isCoordinator = isCoordinator
+    member.save()
+
+    return "", 200
+
+
+@main_bp.route('/members/ban_switch', methods=['POST'])
+def elegibilitySwitch():
```
BrianRamsay: spelling

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+        (SupervisorDepartment.supervisor == supervisorID) &
+        (SupervisorDepartment.department == departmentID)
+    )
+    member.isCoordinator = isCoordinator
+    member.save()
+
+    return "", 200
+
+
+@main_bp.route('/members/ban_switch', methods=['POST'])
```
BrianRamsay: Again with naming - `update_eligibility` or `update_ban`

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+    """
+    Updates a supervisor's eligibility status. 
+    """
+    data = request.get_json()
+    supervisorID = data.get("supervisorID")
+
+    member = Supervisor.get(Supervisor.ID == supervisorID)
+    
+    member.isBanned = not member.isBanned
+    member.save()
```
BrianRamsay, line 160: There's no check in this function to make sure the user is a labor admin. Anyone could update any supervisor, currently.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+    return "", 200
+
+
+@main_bp.route('/members/remove', methods=['DELETE'])
+def removeMember():
+    """
+    Removes a staff member from a department. 
+    """
+    data = request.get_json()
+    supervisorID = data.get("supervisorID")
```
BrianRamsay: Who is allowed to remove a member? There is no authentication.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,172 @@
...
+
+    return "", 200
+
+
+
+@main_bp.route('/members/add', methods=['GET', 'POST'])
+def addUserToDept():
+    """
+    Adds a user to a department.
+    """
```
BrianRamsay: Needs authentication. Who can add a supervisor to a department?

app/controllers/main_routes/main_routes.py
```diff
@@ -65,23 +73,28 @@ def departmentPortal(org=None,account=None):
...
-@main_bp.route('/supervisorPortal/addUserToDept', methods=['GET', 'POST'])
-def addUserToDept():
-    userDeptData = request.form
-    supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = userDeptData['supervisorID'], department = userDeptData['departmentID'])
+                           department = dept,
+                        #    supervisors = supervisor,
+                        #    laborCoordinators=laborCoordinators,
+                           currentUser=g.currentUser,
+                        #    positions = positionsList,
+                        #    posUrl = posUrl
```
BrianRamsay: Let's not commit commented out code

app/controllers/main_routes/main_routes.py
```diff
@@ -65,23 +73,28 @@ def departmentPortal(org=None,account=None):
...
-    
-    except Exception as e:
-        print(f'Could not add user to department: {e}')
-        return "", 500
+        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+    except DoesNotExist:
+        return render_template('errors/404.html'), 404
+
+    positions = Tracy().getPositionsFromDepartment(org, account)
+    print(positions)
```
BrianRamsay: no debug output

app/controllers/main_routes/main_routes.py
```diff
@@ -65,23 +73,28 @@ def departmentPortal(org=None,account=None):
...
-            return "True"
-    
-    except Exception as e:
-        print(f'Could not add user to department: {e}')
-        return "", 500
+        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+    except DoesNotExist:
+        return render_template('errors/404.html'), 404
+
+    positions = Tracy().getPositionsFromDepartment(org, account)
```
BrianRamsay: Are we still getting positions from Tracy? We can use our position table now, right?

app/logic/manageMembers.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def getDepartmentMembers(dept):
+    """Supervisor-department rows for a department, as dicts."""
+    return list(
+        SupervisorDepartment.select(SupervisorDepartment, Supervisor)
+        .where(SupervisorDepartment.department == dept)
+        .join(Supervisor)
+        .dicts()
+    )
```
BrianRamsay: Generally, let's pass around objects, rather than converting them to dicts

app/logic/manageMembers.py
```diff
@@ -0,0 +1,128 @@
...
+            (FormHistory.historyType == "Labor Release Form") &
+            (FormHistory.status == "Approved") &
+            (LaborReleaseForm.releaseDate <= today)
+        )
+    )
+
+
+def getStudentCounts(dept):
+    """Active/pending primary/secondary position counts, keyed by (dept, supervisor)."""
+    releasedForms = getReleasedFormIds()
```
BrianRamsay: if we have a list of IDs, the variable name should reflect that.

BrianRamsay (commented):
Didn't make it through everything, but this should get you started.

app/controllers/main_routes/departmentPortal.py
BhushanSah: Updated this to reuse the existing searchPerson helper from app.logic.search instead of keeping separate name/B-number search logic in this route. This keeps the Manage Members supervisor search aligned with the existing search behavior used elsewhere.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
-from flask import render_template
+from flask import g, jsonify, redirect, render_template, request, url_for
+
+from app.controllers.main_routes import main_bp
+from app.logic.manageMembers import (
```
MImran2002: if the imports are almost all functions we can just do from app.logic.manageMembers import *

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+    attachPositionCounts,
+    canManageMembers,
+    currentAcademicYear,
+    getCurrentDepartment,
+    getDepartmentMembers,
+    getStudentCounts,
+    supervisorsDbToDict,
+)
+from app.logic.search import searchPerson
```
MImran2002, line 11: the imports goes as following sequence controllers, model, logic

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+from app.logic.userInsertFunctions import createSupervisorFromTracy
+from app.models.supervisor import Supervisor
+from app.models.supervisorDepartment import SupervisorDepartment
+
+@main_bp.route('/department/<org>/<account>/members', methods=['GET'])
+def manageMembers(org=None, account=None):
+    """Generates the Manage Members page."""
+    currentUser = g.currentUser
+    
+    if not currentUser.supervisor:
```
MImran2002: this prevents student from managing members but how about other department supervisor, if they can temper with the URL they can access those pages

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+        return redirect(url_for('main.laborhistory', id=currentUser.student.ID))
+
+    dept = getCurrentDepartment(org, account)
```
MImran2002: deciding whether a code should be in logic or a code should be in controller is important here the function getCurrentDepartment can be a single query 
`Department.get(
            Department.ORG == org,
            Department.ACCOUNT == account
        )`
And the return value is use get currentmembers. Those two department functions are small and can be getCurrentDeptMembers.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+
+
+@main_bp.route('/members/search/<query>',  methods=['GET'])
+def searchMember(query=None):
+    """
+    Search supervisors by name or B-number.
+    """
+    currentUser = g.currentUser
+
+    if not canManageMembers(currentUser):
```
MImran2002: this doesn't need to be a function on its own as it checks on authorization.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+    return jsonify(supervisors)
+
+@main_bp.route('/members/update_coordinator', methods=['POST'])
+def updateCoordinator():
+    """
+    Assigns or unassignes a supervisor as a Labor Coordinator. 
+    """
+    currentUser = g.currentUser
+
+    if not canManageMembers(currentUser):
```
MImran2002: the authorization is done through isLaborAdmin and isLaborDepartmentStudent

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+
+
+@main_bp.route('/members/update_eligibility', methods=['POST'])
+def updateEligibility():
+    """
+    Updates a supervisor's eligibility status. 
+    """
+    currentUser = g.currentUser
+
+    if not canManageMembers(currentUser):
```
MImran2002: same here with authorization

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+
+
+@main_bp.route('/members/remove', methods=['DELETE'])
+def removeMember():
+    """
+    Removes a staff member from a department. 
+    """
+    currentUser = g.currentUser
+
+    if not canManageMembers(currentUser):
```
MImran2002: here as well

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+        return render_template('errors/403.html'), 403
+
+    supervisorID = request.form.get("supervisorID")
+    departmentID = request.form.get("departmentID")
+
+    if not supervisorID or not departmentID:
+        return "", 400
+
+    try:
+        supervisorDeptRecord = SupervisorDepartment.get_or_none(
```
MImran2002: if peewee are short make them into one line this means all the peewee command that exist that are multiple line when you can consolidate it into one.

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -1,7 +1,7 @@
 function addSupervisorToDepartment(supervisorID, departmentID, callback=() => {}) {
     return $.ajax({
       method: "POST",
-      url: `/supervisorPortal/addUserToDept`,
+      url: `/members/add`,
```
MImran2002, line 4: if you decide to switch the addusertodept functionality away from supervisorportal you need to also delete the UI build in supervisorportal.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,174 @@
...
+            supervisor=supervisorID,
+            department=departmentID
+        )
+
+        if supervisorDeptRecord:
+            return "False"
+
+        if not Supervisor.get_or_none(Supervisor.ID == supervisorID):
```
MImran2002: good check but we will be omitting the tracy as we don't need to use as the db will be updated weekly

app/controllers/main_routes/main_routes.py
```diff
@@ -54,10 +54,14 @@ def supervisorPortal():
...
+        try:
+            dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+        except (NameError, DoesNotExist):
+            dept = None
+    else:
         dept = None
+    
 
 
     if g.currentUser.isLaborAdmin:
```
MImran2002: do currentUser = g.currentuser.islaboradmin and the current user can be use in return too.

app/logic/manageMembers.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    dbToDict =  {'username': usernameFromEmail(supervisor.EMAIL.strip()),
+                'firstName': supervisor.FIRST_NAME.strip(),
+                'lastName': supervisor.LAST_NAME.strip(),
+                'bnumber': supervisor.ID.strip(),
+                'department': supervisor.DEPT_NAME.strip(),
+                'type': 'Supervisor'}
+    return dbToDict
+
+def currentAcademicYear():
```
MImran2002: this should be better if its turn into global so look to turn this to g.currentYear or g.currentTerm in init

app/logic/manageMembers.py
```diff
@@ -0,0 +1,126 @@
...
+    if today.month < 7: 
+        currentAcademicYear = (currentYear - 1, currentYear)
+    else: 
+        currentAcademicYear = (currentYear, currentYear + 1)
+    # Note that the start of July is 
+    # normally considered the start of a new academic year.
+
+    return currentAcademicYear
+
+def getCurrentDepartment(org, account):
```
MImran2002: these two function can be combine into one.

app/logic/manageMembers.py
```diff
@@ -0,0 +1,126 @@
...
+def getDepartmentMembers(dept):
+    """Supervisor-department rows for a department"""
+    return list(
+        SupervisorDepartment.select(SupervisorDepartment, Supervisor)
+        .join(Supervisor)
+        .where(SupervisorDepartment.department == dept)
+    )
+
+
+def getReleasedFormIds():
```
MImran2002: this can be integrated into getStudentCounts

app/static/css/managemembers.css
```diff
@@ -0,0 +1,162 @@
+/*styles for column alignment in the data table*/
```
MImran2002: manageMembers.css

app/static/css/manageMembers.css
```diff
@@ -0,0 +1,162 @@
+/*styles for column alignment in the data table*/
+.centered-header {
+  text-align: center !important;
```
MImran2002: it seems like AI is used here correct me if I am wrong, there are ways for you to consolidate the selectors and the css too.

app/static/css/managemembers.css
```diff
@@ -0,0 +1,162 @@
...
+  background-color: #FFFFFF;
+  color: #000000;
+}
+
+[class*="hover-indicator"]:hover:after, [class*="hover_indicator"]:focus:after {
+  display: block;
+  color: #000000;
+}
+
+/*a "banned" badge displayed under the name of a banned department member*/
```
MImran2002: remove the comment

app/static/js/manageMembers.js
```diff
@@ -0,0 +1,189 @@
...
+        $.ajax({
+            url: "/members/update_eligibility",
+            data: { supervisorID: supervisorID },
+            type: "POST",
+            success: function() {
+                if (!isBanned) {
+                    button.removeClass("btn-success").addClass("btn-danger");
+                    button.text("Ineligible");
+                    button.val("Ineligible");
+                    ban_badge.css("visibility", "visible");
```
MImran2002: banBadge

app/templates/main/manageMembers.html
```diff
@@ -0,0 +1,141 @@
...
+
+{% block app_content %}
+<h2 class="text-center">{% if department %} Manage {{department.DEPT_NAME}} Members {% else %} Manage Members {% endif %} </h2>
+<h3 class="text-center"> Academic Year: {{currentAcademicYear[0]}}–{{currentAcademicYear[1]}}</h3>
+{%if g.currentUser.supervisor.isBanned %}
+
+<br></br>
+<div class = "alert alert-danger" role="alert"> Your account is currently ineligible to add new students or create new status forms. Please contact the Labor Office for more information.
+ </div> 
+ <script>
```
MImran2002: we don't use scripts in html when it comes to logic

database/demo_data.py
```diff
@@ -743,6 +936,42 @@
     ]
 Allocation.insert_many(allocations).on_conflict_replace().execute()
 
+print("Data insertion complete :)")
```
MImran2002: remove this

tests/code/test_departmentPortal.py
```diff
@@ -0,0 +1,942 @@
+import pytest
+
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+from flask import Flask, g
+
+import app.controllers.main_routes.departmentPortal as department_portal
+
+
+@pytest.fixture
```
MImran2002: genuine question what is your reason behind creating app?

tests/code/test_departmentPortal.py
```diff
@@ -0,0 +1,942 @@
...
+def fakeLaborDepartmentStudentUser():
+    return SimpleNamespace(
+        supervisor=SimpleNamespace(ID="B00000098"),
+        student=None,
+        isLaborAdmin=False,
+        isLaborDepartmentStudent=True,
+    )
+
+
+def fakeSupervisorUser():
```
MImran2002: why can't we create this data as peewee create instead of function?

tests/code/test_departmentPortal.py
```diff
@@ -0,0 +1,942 @@
...
+    )
+
+
+# -------------------------------------------------------------------
+# manageMembers tests
+# -------------------------------------------------------------------
+
+
+@pytest.mark.integration
+def test_manage_members_for_supervisor(client, monkeypatch):
```
MImran2002: naming convention

tests/code/test_departmentPortal.py
```diff
@@ -0,0 +1,942 @@
...
+
+
+# -------------------------------------------------------------------
+# searchMember tests
+# -------------------------------------------------------------------
+
+
+@pytest.mark.integration
+def test_search_member_by_bnumber(client, monkeypatch):
+    fake_user = fakeAdminUser()
```
MImran2002: please refactor the test file first study the the existing test file and create them test similar to those structure.

MImran2002 (changes requested):

app/static/css/manageMembers.css
BhushanSah: I am not sure about this as this was already written by the previous team.

app/static/js/manageMembers.js
```diff
@@ -0,0 +1,190 @@
...
+        $.ajax({
+            url: "/members/update_coordinator",
+            data: { supervisorID: supervisorID, departmentID: departmentID, isCoordinator: isChecked },
+            type: "POST",
+            success: function() {
+                if (isChecked) {
+                    $("#flash_container").html("<div class=\"alert alert-success\" role=\"alert\" id=\"flasher\">" + memberName + " has been assigned as a coordinator.</div>");
+                    $("#flasher").delay(3000).fadeOut();
+                } else {
+                    $("#flash_container").html("<div class=\"alert alert-info\" role=\"alert\" id=\"flasher\">" + memberName + " is no longer a coordinator.</div>");
```
conwelld: Any removal should display a red banner instead of a green one to indicate the difference from adding members to a department.

conwelld (changes requested):
I think the color of the pop-up that confirms a member's removal should change, because it is not consistent with the color used for warnings.

app/controllers/main_routes/__init__.py
```diff
@@ -1,19 +1,32 @@
...
 from flask import render_template
 from flask import Blueprint
 from app.login_manager import require_login
 import os
 
 main_bp = Blueprint('main', __name__)
+
+def currentAcademicYear():
+    today = date.today()
+    currentYear = today.year
```
MImran2002: this code should be in app>init.py and function name should be change too and referred to celts repo to see the changes

app/controllers/main_routes/__init__.py
```diff
@@ -1,19 +1,32 @@
...
+
+    if today.month < 7:
```
MImran2002: using in global current year will be more valuable than g.currentacademicyear

app/controllers/main_routes/departmentPortal.py
BhushanSah: I updated the access check so students are still redirected, but supervisors can only access the Manage Members page for departments they are actually assigned to.

The page now allows labor admins, labor department student workers, and supervisors connected to that specific department through SupervisorDepartment. Supervisors from other departments should get a 403 if they try to change the URL.

MImran2002 (changes requested):

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1 +1,177 @@
...
+    return jsonify(supervisors)
+
+@main_bp.route('/members/update_coordinator', methods=['POST'])
+def updateCoordinator():
+    """
+    Assigns or unassignes a supervisor as a Labor Coordinator. 
+    """
+    currentUser = g.currentUser
+
+    if not (currentUser.isLaborAdmin or currentUser.isLaborDepartmentStudent):
```
MImran2002: This should be able to be done by supervisors or labor coordinators too in the department so you need to give them access too.

app/logic/manageMembers.py
```diff
@@ -0,0 +1,115 @@
...
+from app.logic.search import usernameFromEmail
+from app.models.department import Department
+from app.models.formHistory import FormHistory
+from app.models.laborReleaseForm import LaborReleaseForm
+from app.models.laborStatusForm import LaborStatusForm
+from app.models.supervisor import Supervisor
+from app.models.supervisorDepartment import SupervisorDepartment
+
+
+def supervisorsDbToDict(supervisor):
```
MImran2002: what you should do is remove this function totally in app>logic>getsupervisors.py there is a function call buildSupervisorDisplay that is technically what you are doing and that function is there but it is smaller go ahead and make changes to it so that it reflects what you want to but at the same time there is a route that uses the buildsupervisordisplay function so make the test for the logic and controller are not broken and are returning right

app/logic/manageMembers.py
```diff
@@ -0,0 +1,115 @@
...
+        'firstName': supervisor.FIRST_NAME.strip(),
+        'lastName': supervisor.LAST_NAME.strip(),
+        'bnumber': supervisor.ID.strip(),
+        'department': supervisor.DEPT_NAME.strip(),
+        'type': 'Supervisor'
+    }
+    return dbToDict
+
+
+def getCurrentDeptMembers(org, account):
```
MImran2002: the same in the app>logic>getSupervisors in getSupervisors function your getCurrentDeptmembers and getSupervisor function technically do something similar combine them make sure the getSupervisor test and logic is not breaking and make sure the controller that calls that logic is also working.

app/logic/manageMembers.py
```diff
@@ -0,0 +1,115 @@
...
+        SupervisorDepartment
+        .select(SupervisorDepartment, Supervisor)
+        .join(Supervisor)
+        .where(SupervisorDepartment.department == dept)
+    )
+
+    return dept, members
+
+
+def getStudentCounts(dept):
```
MImran2002: for this talk to @fritzj2 @BetterRocket and @nahom70  they are working on a function for this and they have the correct query combination ask them about it and use it.

app/static/js/addSupervisorsToDepartment.js
```diff
@@ -1,26 +1,27 @@
 function addSupervisorToDepartment(supervisorID, departmentID, callback=() => {}) {
```
MImran2002, line 1: if addsupervisorToDeparment is only for managemember page we can definitely move this whole code into managemembers.js file

tests/code/test_departmentPortal.py
```diff
@@ -0,0 +1,228 @@
...
+            isSaasAdmin=False
+        )
+
+        SupervisorDepartment.create(
+            supervisor=memberSupervisor,
+            department=dept,
+            isCoordinator=False
+        )
+
+        with app.test_request_context(
```
MImran2002: you do not create any logic for the deparmentportal logic so this can be removed

tests/code/test_manageMembers.py
```diff
@@ -0,0 +1,234 @@
...
+        assert members[0].active_primary_positions == 0
+        assert members[0].pending_primary_positions == 0
+        assert members[0].active_secondary_positions == 0
+        assert members[0].pending_secondary_positions == 0
+
+        transaction.rollback()
+
+
+@pytest.mark.integration
+def test_getStudentCounts_counts_positions_and_excludes_released_forms():
```
MImran2002: if this test is to test getstudentcount fucniton it should be name test_getStudentCount and tell a story that lead to testing the function.

MImran2002 (changes requested):

tests/code/test_departmentPortal.py
BhushanSah: I removed the large combined departmentPortal test and kept the tests focused.

tests/code/test_manageMembers.py
BhushanSah: I also renamed the getStudentCounts test and added a clearer Given/When/Then flow so the test explains the setup: one department, one supervisor, active and pending primary/secondary forms, plus one released form that should be ignored.

app/static/js/addSupervisorsToDepartment.js
brightfietsop-ux: This helper is also used by manageDepartments.js and is loaded by admin/manageDepartments.html, so it is not exclusive to the Manage Members page. I kept it as a shared JavaScript helper to avoid duplicating the same add-supervisor request logic.

app/static/css/manageMembers.css
```diff
@@ -0,0 +1,150 @@
...
```
MImran2002: Image make sure the width of this fits the screen

MImran2002: like edit the div that contains the data table so that it stays within the screen cause now i have to scroll side way

app/static/css/manageMembers.css
```diff
@@ -0,0 +1,150 @@
...
+.centered-cell {
+  vertical-align: middle !important;
+}
+
+#add-member {
+  display: flex;
+  justify-content: center;
+  margin-bottom: 20px;
+  z-index: 100;
+  gap: 0;
```
MImran2002: Image buttons are asymmetric height wise

MImran2002 (changes requested):
As you can see here I created a form under wario nakazawa 

Image there is no approve form but it is showing active here 

Image eventhough wario has no approved form yet. 

But when i look there is a approved secondary for Elehah for secondary so that might influence the result. Moreover, there is a problem where 

image as you can see scott has active secondary but in the manage member it only shows pending.

MImran2002 (commented):
eventhough it says 2026-2027 the active and pending that are there shows those of 2025-2026

Image

MImran2002 (changes requested):
eventhough it says 2026-2027 the active and pending that are there shows those of 2025-2026

Image

app/controllers/main_routes/__init__.py
```diff
@@ -1,3 +1,4 @@
+from datetime import date
```
MImran2002: remove this import statement

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,28 +1,244 @@
...
+
+@main_bp.route('/department/<org>/<account>/members', methods=['GET'])
+def manageMembers(org=None, account=None):
+    """Generates the Manage Members page."""
+    currentUser = g.currentUser
+
+    if not currentUser.supervisor:
+        return redirect(url_for('main.laborhistory', id=currentUser.student.ID))
+
+    dept = Department.get_or_none(
```
MImran2002: dept = Department.get_or_none(Department.ORG == org, Department.ACCOUNT == account) one line peewee command

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,28 +1,244 @@
...
+        Department.ORG == org,
+        Department.ACCOUNT == account
+    )
+
+    if not dept:
+        abort(404)
+
+    members = getSupervisorDepartments(dept)
+
+    supervisorDeptRecord = SupervisorDepartment.get_or_none(
```
MImran2002: single line peewee command

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,28 +1,244 @@
...
+    )
+
+    if not (
+        currentUser.isLaborAdmin or
+        currentUser.isLaborDepartmentStudent or
+        supervisorDeptRecord
+    ):
+        return render_template('errors/403.html'), 403
+
+    counts = getStudentCounts(dept)
```
MImran2002: name a variable that actually describe the output of the function for both counts and members

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,28 +1,244 @@
...
+    members = attachPositionCounts(members, counts)
+
+    return render_template(
+        'main/manageMembers.html',
+        members=members,
+        department=dept,
+    )
+
+
 @main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
```
MImran2002, line 74: why are you touching the position route? this is outside of your scope

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,28 +1,244 @@
...
+            department=departmentID
+        )
+
+        if supervisorDeptRecord:
+            return jsonify(
+                success=False,
+                message="Supervisor already exists in this department."
+            ), 200
+
+        SupervisorDepartment.create(
```
MImran2002: same here

app/logic/manageMembers.py
```diff
@@ -0,0 +1,101 @@
...
+
+from peewee import Case, fn
+
+from app.models.formHistory import FormHistory
+from app.models.laborReleaseForm import LaborReleaseForm
+from app.models.laborStatusForm import LaborStatusForm
+from app.models.term import Term
+
+
+def getStudentCounts(dept):
```
MImran2002: rename the function your function description and function name seems to be different, function names should give a hint of what the function do i believe the function here returns active pending primary positions no??

app/__init__.py
```diff
@@ -79,9 +79,21 @@ def load_openTerm():
             session['openTerm'] = model_to_dict(term)
         g.openTerm = term
         
+def currentYear():
+    today = date.today()
```
MImran2002, line 93: this should be g.currentYear you will have to make this global

app/__init__.py
```diff
@@ -79,9 +79,21 @@ def load_openTerm():
...
+        return year - 1, year
+
+    return year, year + 1
+        
 @app.context_processor
 def inject_environment():
-    return dict(env=app.config['ENV'])
+    return dict(
+        env=app.config['ENV'],
+        currentYear=currentYear()
```
MImran2002: why do you need to inject this?

tests/code/test_departmentPortal.py
```diff
@@ -0,0 +1,110 @@
...
+
+        assert data[0]["bnumber"] == "B99000001"
+        assert data[0]["firstName"] == "Test"
+        assert data[0]["lastName"] == "Supervisor"
+
+        transaction.rollback()
+
+
+@pytest.mark.integration
+def test_addUserToDept_adds_existing_supervisor():
```
MImran2002: should be just test_addUserToDept as we are moving and another thing as adduserToDept basically move from one file to another make sure only the test for addusertodept is consider under test_departmentportal.py

tests/code/test_manageMembers.py
```diff
@@ -0,0 +1,234 @@
...
+        members = getSupervisorDepartments(department)
+
+        assert len(members) == 1
+        assert members[0].supervisor.ID == supervisor.ID
+        assert members[0].department.departmentID == department.departmentID
+
+        transaction.rollback()
+        
+@pytest.mark.integration
+def test_buildSupervisorDisplay_returns_portal_and_search_fields():
```
MImran2002: should only be test_buildSupervisorDisplay as its name

tests/code/test_manageMembers.py
```diff
@@ -0,0 +1,234 @@
...
+        assert members[0].pending_primary_positions == 0
+        assert members[0].active_secondary_positions == 0
+        assert members[0].pending_secondary_positions == 0
+
+        transaction.rollback()
+
+
+@pytest.mark.integration
+@pytest.mark.integration
+def test_getStudentCounts_counts_active_pending_positions_and_ignores_released_form():
```
MImran2002: same here we should only have one test for one function and the format just should be test_functionname and for the scenarios that should be considered within that test function it shouldn't have its own separate test function

tests/code/test_manageMembers.py
```diff
@@ -0,0 +1,234 @@
...
+
+        term = Term.get_or_create(
+            termCode=202600,
+            defaults={
+                "termName": "AY 2026-2027",
+                "termStart": "2026-08-01",
+                "termEnd": "2027-05-01",
+                "termState": 1,
+                "primaryCutOff": "2026-09-01",
+                "adjustmentCutOff": "2026-10-01"
```
MImran2002, line 174: and the test are broken go through each one of em and fix, look into the logic don't just change the output please

MImran2002 (changes requested):

app/controllers/main_routes/__init__.py
BhushanSah: Done

app/controllers/main_routes/departmentPortal.py
BhushanSah: Done

app/controllers/main_routes/departmentPortal.py
BhushanSah: Done

app/__init__.py
BhushanSah: Made this

app/logic/manageMembers.py
BhushanSah: renamed

app/controllers/main_routes/departmentPortal.py
BhushanSah: renamed

app/controllers/main_routes/departmentPortal.py
BhushanSah: fixed this too

app/controllers/main_routes/departmentPortal.py
BhushanSah: done

tests/code/test_manageMembers.py
BhushanSah: I refactored the tests based on the requested structure. Each tested function now has one test named test_functionName, and the different scenarios are handled inside that test instead of creating separate scenario-specific test names. I also fixed the broken getActivePendingPositionCounts test by matching the function call to the current helper signature and adding FormHistory rows for the Labor Status Form scenarios.

app/__init__.py
BhushanSah: Updated this to follow the same pattern as g.currentUser. Instead of injecting currentYear through the context processor, I now load it during before_request and access it as g.currentYear in the template. The context processor only keeps env now.

fritzj2:
Currently, this is the implementation of the manage members page.

Why is it 15 files?
Primarily, this is due to either small file changes, such as `__init__.py` needing 1 additional line. But this also covers a change that requires moving one file to another. The addUserToDept() function got moved from the supervisor portal to the members page. This caused around 4 files to be changed just to remove it from the page and reroute it to the members page.

This PR also has a bug fix baked into it regarding the stale user session. This patches the issue where, when switching users, the cookie would persist and not change out the session unless the cache was cleared. I do not know why this is in this issue.

Then, the actual implementation of the issue concerns the rest of the files. The code is now in parity with the rest of the changes and uses functions created after the PR was submitted. The code has also been trimmed significantly to the best of my ability without breaking the files.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -88,3 +95,163 @@ def managePositions(org, account):
...
+    currentUser = g.currentUser
+
+    if not (currentUser.isLaborAdmin or currentUser.isLaborDepartmentStudent or currentUser.supervisor ):
+        return render_template('errors/403.html'), 403
+
+    supervisors = (
+        searchPerson(Supervisor, query)
+        .order_by(Supervisor.LAST_NAME.asc())
+        .limit(10)
+    )
```
ArtemKurasov: I think that this is unnecessarily split into 5 lines. The query is not so long that it can't be on one line.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -88,3 +95,163 @@ def managePositions(org, account):
...
                            positions = positions
                            )
+
+
+@main_bp.route('/members/search/<query>', methods=['GET'])
+def searchMember(query=None):
+    """Search supervisors by name or B-number."""
+    currentUser = g.currentUser
+
+    if not (currentUser.isLaborAdmin or currentUser.isLaborDepartmentStudent or currentUser.supervisor ):
```
ArtemKurasov: Could you remove the space before the closing parenthesis?

app/controllers/main_routes/departmentPortal.py
```diff
@@ -88,3 +95,163 @@ def managePositions(org, account):
...
+
+    if not supervisorID or not departmentID:
+        return "", 400
+
+    supervisorDeptRecord = None
+
+    if currentUser.supervisor:
+        supervisorDeptRecord = SupervisorDepartment.get_or_none( (SupervisorDepartment.supervisor == currentUser.supervisor) & (SupervisorDepartment.department == departmentID))
+
+    if not (currentUser.isLaborAdmin or currentUser.isLaborDepartmentStudent or supervisorDeptRecord ):
```
ArtemKurasov: Can you remove the space before the closing parenthesis in the condition?

app/controllers/main_routes/departmentPortal.py
```diff
@@ -88,3 +95,163 @@ def managePositions(org, account):
...
+    departmentID = request.form.get("departmentID")
+    isCoordinator = request.form.get("isCoordinator") == "true"
+
+    if not supervisorID or not departmentID:
+        return "", 400
+
+    supervisorDeptRecord = None
+
+    if currentUser.supervisor:
+        supervisorDeptRecord = SupervisorDepartment.get_or_none( (SupervisorDepartment.supervisor == currentUser.supervisor) & (SupervisorDepartment.department == departmentID))
```
ArtemKurasov: Could you remove the space after the opening parenthesis and before `(SupervisorDepartment`?

app/controllers/main_routes/departmentPortal.py
```diff
@@ -88,3 +95,163 @@ def managePositions(org, account):
...
+
+    return "", 200
+
+
+@main_bp.route('/members/update_eligibility', methods=['POST'])
+def updateEligibility():
+    """Updates a supervisor's eligibility status."""
+    currentUser = g.currentUser
+
+    if not ( currentUser.isLaborAdmin or currentUser.isLaborDepartmentStudent or currentUser.supervisor):
```
ArtemKurasov: Could you remove the space after the opening parenthesis?

app/logic/getSupervisors.py
```diff
@@ -1,43 +1,65 @@
...
+        "bnumber": supervisor.ID.strip(),
+        "department": (supervisor.DEPT_NAME or "").strip(),
+        "type": "Supervisor"
     }
-def getSupervisors(dept):    
+
+
+def getSupervisorDepartments(dept):
+    """Return supervisor-department records for a department."""
+    if dept is None:
```
ArtemKurasov: Can the value of `dept` be anything other than a Department object or None? If not, I would write line 33 as: 

```python
if not dept:
``` 

_**Note:** Writing `if dept is None` can be helpful if `dept` can also be equal to `False` or `0`, but I doubt it's the case for this particular variable._

app/logic/getSupervisors.py
```diff
@@ -1,43 +1,65 @@
...
-
-            if supervisorDepartment.isCoordinator:
-                laborCoordinators.append(supervisorDisplay)
-            else:
-                supervisors.append(supervisorDisplay)
+
+    for supervisorDepartment in getSupervisorDepartments(dept):
+        supervisor = supervisorDepartment.supervisor
+
+        if supervisor is None:
```
ArtemKurasov: Again, if supervisor cannot be anything other than a Supervisor object or `None`, I would write the condition on line 52 as 

```python
if not supervisor:
```

What's more, on line 57, we can see the same style of code as I am suggesting.

tests/code/test_manageMembers.py
```diff
@@ -0,0 +1,247 @@
+from datetime import date
+
+import pytest
+
+from app.logic.getSupervisors import buildSupervisorDisplay, getSupervisorDepartments
+from app.logic.manageMembers import (
+    attachPositionCounts,
+    getActivePendingPositionCounts,
+)
```
ArtemKurasov: Can we just write this import statement in one line? 

```python
from app.logic.manageMembers import attachPositionCounts, getActivePendingPositionCounts
```

app/static/js/manageMembers.js
```diff
@@ -0,0 +1,177 @@
...
+
+    // wait a little longer for bnumber typing
+    let keyInterval = 200
+    if (e.keyCode >= 48 && e.keyCode <= 57) {
+        keyInterval = 500
+    }
+
+    // don't search for every key (especially relevant for bnumber)
+    clearTimeout(typeTimer)
+    typeTimer = setTimeout(function() { sendQuery(e.target.value); }, keyInterval)
```
ArtemKurasov: Why do we need a delay for B-numbers?

ArtemKurasov (commented):
I added some comments to this PR, suggesting some minor changes, but overall, everything is well done!

PR #652, opened 2026-07-20
Add a Position card (fixed) into department Portal

fritzj2:
## Issue Description

Fixes issue #604 
- The department portal needs to have a summary view of the positions that each department has.
- This should include only a small subset of the positions
- This should display the WLS and position Title.

## Additions

image 
image

- Added 3 Cards to the manage department page
- The position card has a list of `positionHistory Positions` and their WLS level
- The list of positions takes you to a link where the specific position info should be (but currently isn't implemented)
- The "View all positions" button takes you to where the " Manage/View Positions page will lead.
- Failsafe for no positions in a department ("No positions for this department" appears instead)
- Failsafe for an overload of positions (and x amount more positions)
- `departmentPortal.css` handles all of the styling for the cards.
- added `getPositions.py` to handle the logic of getting positions

## Additions (7/22/2026)
- Added test suite (test_getPosition.py).
- Updated the postitionHistory model to no longer enforce a uniqueness constraint that forced all positions to have a unique revisionDate.
- Updated getPositions to now account for duplicate position codes that go to different departments. (Duplicates can not exist across departments are each position key is unique if active. (Duplicates are fine for inactive positions for history purposes).)

## Changes

- Department portal Department selector dropdown no longer takes up the entire screen
- Cards have a variable size that scales with the screen
- The cards stack on top of each other when the screen size is too small.
- The cards do _not_ appear if a department has not been selected.
- `migrate_db.sh` contains an extra line to accommodate the positionHistory being added to the table.
- `positionHistory.py` contains 1 extra line that now contains the `positionTitle`
- Additional files can be found from a development PR merge, which has not yet been implemented into department-portal-base.
- `posUrl` has been changed to `posURL`
- The check for no positions in a department has been changed.
  - It no longer uses a if list is empty -> set list = specific string -> check for specific string
  - Now checks if the list is empty -> pass check if the list is empty
    - If the list is empty, it will say No active positions and move on
    - If the list has elements, it will print out up to 7 elements, then check if it is over, and print out the "and x more"
## Testing

- Reset the database to include the changes to the `demo_data.py`.
- Booting into the portal using flask run allows you to see if the card is there.
- 10 positions should load for the CS department ( 7 + 3 more), while 1 should load for the Mathematics department
  - The mathematics department has 2 positions assigned to it; however, only 1 is active, so only one loads.
- The demo_data.py file has been updated to include `positionHistory` data within it.
  - The position info page does not yet exist and will return a 404 error when selected; however, the URL still does exist.
- The "View All Positions" button will give you a 404 error when selected, as the webpage for it doesn't exist yet; the URL still exists.
- Mess with the screen size to ensure that the cards do not have a static size. The 2 extra cards will go side by side when the screen is small enough.
- Ensure that the largest size of the department doesn't extend out of the dropdown bar
- This data will not load anything on staging data currently, as nothing has been allocated to positionHistory yet.
- Use the test suite.


app/templates/main/departmentPortal.html
```diff
@@ -24,4 +25,70 @@ <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% e
...
+        <section class="card" aria-labelledby="positions-title">
+          <div class="card-body">
+              <div class="media">
+                <div class="media-left media-middle">
+                  <span aria-hidden="true" style="color: #424242; font-size: 24px;">
+                    <i class="bi bi-suitcase-lg-fill"></i>
+                  </span>
+                </div>
+              <div class="media-body media-middle">
+                <h1>Positions</h1>
```
MImran2002: `<h3>`

app/templates/main/departmentPortal.html
```diff
@@ -24,4 +25,70 @@ <h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Portal {% e
...
+                  <p> No active positions in this department </p>
+                </li>
+              {% else %}
+                {% for p in positions[:7] %}
+                    <li class="card-text">
+                      <a href= "/department/{{ department.ORG }}/{{ department.ACCOUNT }}/positions/{{posURL[loop.index0]}}">{{ p }}</a>
+                    </li>
+                {% endfor %}
+                {% if (positions | length) > 7 %}
+                  </ul><p class="card-text">and {{ (positions | length) - 7}} more...</p>
```
MImran2002: more ...

MImran2002 (changes requested):
needed to create test

app/logic/getPositions.py
```diff
@@ -0,0 +1,45 @@
...
+    seenCodes = set()
+    # Iterate through the positions and filter out duplicates based on positionCode and department
+    for pos in positions:
+        code = pos.positionCode
+
+        # Same dept returned two Active rows with the same code (shouldn't normally happen, but guards against bad data) -> skip the duplicate
+        if code in seenCodes:
+            continue
+
+        # Enforces: A positionCode can only be Active in ONE place at a time, regardless of department or revisionDate.
```
MImran2002: remove all those comments

app/logic/getPositions.py
```diff
@@ -0,0 +1,45 @@
...
+    if not dept:
+        return [], []
+
+    positions = list(PositionHistory.select()
+                    .where(PositionHistory.department == dept, PositionHistory.status == "Active")
+                    .order_by(PositionHistory.positionTitle.asc()))
+
+    positionsList = []
+    posURL = []
+    seenCodes = set()
```
MImran2002: seenPositionCodes

app/logic/getPositions.py
```diff
@@ -0,0 +1,45 @@
...
+    # Iterate through the positions and filter out duplicates based on positionCode and department
+    for pos in positions:
+        code = pos.positionCode
```
MImran2002: positionCode or something not just code as a var name

app/logic/getPositions.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # Same dept returned two Active rows with the same code (shouldn't normally happen, but guards against bad data) -> skip the duplicate
+        if code in seenCodes:
+            continue
+
+        # Enforces: A positionCode can only be Active in ONE place at a time, regardless of department or revisionDate. 
+        # If an earlier-created Active row with this code exists in a DIFFERENT department, this one loses.
+        claimedElsewhere = (PositionHistory
```
MImran2002: isPositionClaimed

MImran2002 (changes requested):

app/models/positionHistory.py
```diff
@@ -10,5 +11,5 @@ class PositionHistory(baseModel):
     description        = TextField(default=None)
 
     class Meta:
-        indexes = ( (('positionCode', 'revisionDate', 'status'), True), )
+        indexes = ( (('positionCode', 'revisionDate', 'status'), False), )
```
MImran2002: I believe they must be true because lets say the position is student programmer and its S1234 and the revision date is 2/3/2026 and status is pending. if we have it false that mean we cannot have one uniqueness at all. we should allow uniqueness so that there can be the same active status, on the same revision date, positioncode twice

app/logic/getPositions.py
```diff
@@ -0,0 +1,41 @@
...
+        if positioncode in seenPositionCodes:
+            continue
+
+        isPositionClaimed = (PositionHistory
+                            .select()
+                            .where(
+                                PositionHistory.positionCode == positioncode,
+                                PositionHistory.status == "Active",
+                                PositionHistory.department != dept,
+                                PositionHistory.id < pos.id
```
MImran2002: talked to brian and scott we can remove the duplicate check which means we can also remove is positionclaimed: continue too.

app/logic/getPositions.py
```diff
@@ -0,0 +1,41 @@
...
+    if not dept:
+        return [], []
+
+    positions = list(PositionHistory.select()
+                    .where(PositionHistory.department == dept, PositionHistory.status == "Active")
+                    .order_by(PositionHistory.positionTitle.asc()))
+
+    positionsList = []
+    posURL = []
+    seenPositionCodes = set()
```
MImran2002: this shouldn't be seenpositioncode now i think of it you are not seeing it this is a set to hold unique positioncode so name something the data structure holds

tests/code/test_getPositions.py
```diff
@@ -0,0 +1,385 @@
...
+
+        # Check if position list is returned in alphabetical order
+        assert positionsList[0] == "Intern: (WLS 1)"
+        assert positionsList[1] == "Lab Assistant: (WLS 2)"
+        assert positionsList[2] == "Teaching Assistant: (WLS 4)"
+
+        # Check if the inactive position is not included in the results
+        assert "Research Assistant: (WLS 3)" not in positionsList
+
+        # Check if posURL contains the correct position codes in the same order as positionsList
```
MImran2002, line 66: most of our test focus on functions so right now your test functions are a reflection of different scenarios instead of different function which means if you only work on one function which is the getActivePosition that should be reflected there. So your other test can be narrowed down actually.

tests/code/test_getPositions.py
```diff
@@ -0,0 +1,385 @@
...
+        positionsList, posURL = getActivePositions(None)
+
+        # Check that no positions are returned when department is None
+        assert len(positionsList) == 0
+        assert len(posURL) == 0
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_checkPositionDuplicates():
```
MImran2002: positions do not duplicate we can ignore this scenario

tests/code/test_getPositions.py
```diff
@@ -0,0 +1,385 @@
...
+
+        # Check that the correct position titles and codes are returned for department B and A (Focused on the contested position code "S34522" due to the duplicate across departments) 
+        assert "Research Assistant: (WLS 3)" in positionsListB
+        assert "Book Handler: (WLS 4)" not in positionsListB
+        assert "Book Handler: (WLS 4)" not in positionsListA
+        assert "S34523" in posURLB
+        assert "S34522" not in posURLB
+        transaction.rollback()
+
+@pytest.mark.integration
```
MImran2002: this should also be removed as position code is unique so we won't face a order problem.

tests/code/test_getPositions.py
```diff
@@ -0,0 +1,385 @@
...
+
+        # deptB claimed the code first therefore deptA should lose it, not deptB
+        assert len(positionsListA) == 0
+        assert len(positionsListB) == 1
+        assert "Archivist: (WLS 2)" in positionsListB
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_checkDuplicateActiveWithinSameDepartment():
```
MImran2002: this can be avoided if we allow index to be true then it must meet the high threshold of revisiondate, positioncode and status being different and whenever we even edit the position we are changing the revision date not creating a new row.

tests/code/test_getPositions.py
```diff
@@ -0,0 +1,385 @@
...
+        positionsList, posURL = getActivePositions(dept)
+
+        # Only one should be counted despite two Active rows with the same code
+        assert len(positionsList) == 1
+        assert len(posURL) == 1
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_checkThreeWayPositionCodeConflict():
```
MImran2002: this scenario will not happen as position codes are unique

tests/code/test_getPositions.py
```diff
@@ -0,0 +1,385 @@
...
+        positionListC, posUrlC = getActivePositions(deptC)
+
+        assert len(positionListA) == 1  # only the first-created claim survives
+        assert len(positionListB) == 0
+        assert len(positionListC) == 0
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_checkInactiveElsewhereDoesNotBlockActiveClaim():
```
MImran2002: this scenario is interesting but we are still basing it off the idea that positioncode are unique

MImran2002 (changes requested):

app/static/css/departmentPortal.css
```diff
@@ -0,0 +1,31 @@
...
+.form-group {
+    width: 50%;
+    margin-left: auto ;
+    margin-right:auto;
+}
+.card-body {
+  padding: 1rem 1rem;
+  min-width: 100%
+}
+.row {
```
MImran2002: so what is happening here is you are creating custom css class row while bootstrap class row also have its own css so my advise is to rename the class to card-row or something and use that and what you will see is it becomes more responsive and when those tables are empty assign a minimum height with css so that they don't get squished.

app/static/css/departmentPortal.css
```diff
@@ -0,0 +1,31 @@
+.card {
+  border-radius: 1rem;
+  overflow: hidden;
+  width: 100%;
+  height: 100%;
+  min-width: 100%;
+  box-shadow: 2px 2px 4px rgba(100, 100, 100, 0.26);
```
MImran2002, line 7: add   margin-bottom: 20px; here to separate the card when the cards are stacekd

tests/code/test_getPositions.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+from app.models import mainDB
+from app.models.department import Department
+from app.models.positionHistory import PositionHistory
+from app.logic.getPositions import getActivePositions
+
+@pytest.mark.integration
+def test_getPositions():
```
MImran2002: test_getActivePositions

MImran2002 (changes requested):
change the css that i recommended the test looks good except the test function name

MImran2002 (approved):
all looks good

PR #655, opened 2026-07-21
Removed --skip-ssl from reset_database.sh

BetterRocket:
## Issue Description
`--skip-ssl` is present in department-portal-base, but no one is using dev containers anymore so `--skip-ssl` now causes an sql alchemy error. It's now deprecated in the newer mysql version that runs in the non-container WSL setup, and is entirely redundant.

Fixes #add issue number
Making the issue would take longer than fixing the problem. 

## Changes

- Removed every instance of `--skip-ssl` from `reset_database.sh`

## Testing

- Ensure no other files were changed and that the only change is in `reset_database.sh`
- Ensure that the non-container WSL setup does not require `--skip-ssl`

### DIFF
--- a/database/reset_database.sh
+++ b/database/reset_database.sh
@@ -22,12 +22,12 @@ if [ "$1" == "from-backup" ]; then
 fi
 
 echo "Dropping databases"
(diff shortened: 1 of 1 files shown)


ArtemKurasov (commented):
No one uses dev containers, so these changes should be fine

BrianRamsay (approved):

PR #656, opened 2026-07-21
Manage Positions Page 

NYABUTOA:
Fixes issue https://github.com/BCStudentSoftwareDevTeam/lsf/issues/644

**Changes**

1. Created a position route in the main route.
2. Queried data from the positions history to filter the active positions in history and pass to managepositions.html
3. Edited demo data for position history, tweaked the attributes there to match the model.
4. Added managepositions.css, where some CSS logic is handled
5. Added managepositions.js, where the datatable logic is handled

**Testing**

Worked around the data and UI through inspection
Made sure the passed data is displayed by cross-checking on the database.


app/controllers/main_routes/main_routes.py
```diff
@@ -51,37 +57,129 @@ def supervisorPortal():
...
+                           break_hours = sumBreak,
+                           positions = positionsList,
+                           posUrl = posUrl,
+                           supervisors = supervisors,
+                           laborCoordinators=laborCoordinators,
+                           currentUser=g.currentUser
+                           )
+
+
+@main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
```
BetterRocket: duplicated, remove one

app/controllers/main_routes/main_routes.py
```diff
@@ -51,37 +57,129 @@ def supervisorPortal():
...
+@main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
+def managePositions(org, account):
```
BetterRocket, line 164: this code doesn't check if the current user can access this specific department (aka can type whatever into URL and see whatever)

database/demo_data.py
```diff
@@ -752,48 +1013,202 @@
...
+    {
+        "positionTitle": "Teaching Associate",
         "positionCode": "S61410",
+        "department": 1,
         "status": "Inactive",
         "wls":2,
-        "revisionDate" : f"2025-01-01",
+        "revisionDate" : f"2026-01-01",
+        "description": "",
+        "department" : 3
```
BetterRocket: why two department fields?
This is already in `department-portal-base` but you might as well fix it.

BetterRocket (changes requested):

app/controllers/main_routes/main_routes.py
NYABUTOA: lines 164 through 166 are meant to check if the user is supposed be able to enetr that portal or not.

PR #658, opened 2026-07-22
Manage positions page implementation

conwelld:
Fixes issue https://github.com/BCStudentSoftwareDevTeam/lsf/issues/644

Changes

Created a position route in the main route.
Queried data from the positions history to filter the active positions in history and pass to managepositions.html
Added managepositions.css, where some CSS logic is handled
Added managepositions.js, where the datatable logic is handled
Added logic to preserve allocation privacy
image
image

Testing

It is important to note that cookies for this website will retain the user's role as long as you run it; clear the cookies if you change the user's role. This will make it so that a user should only see the departments they should access when changing the department ID
Worked around the data and UI through inspection
Made sure the passed data is displayed by cross-checking on the database.
changed it to if a user changes the url to another department, th euser can not see other department's allocations if they are not a labor admin

### DIFF
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -74,15 +74,17 @@ def departmentPortal(org=None,account=None):
                            department = dept,
                            positions = positionsList,
                            posURL = posURL)
+
 @main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
 def managePositions(org, account):
     try:
         dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
     except DoesNotExist:
         return render_template('errors/404.html'), 404
-
+    
     if not g.currentUser.isLaborAdmin:
-        allowedDepartmentIds = [d.departmentID for d in getDepartmentsForSupervisor(g.currentUser)]
+        supervisorDepts = SupervisorDepartment.select().where(SupervisorDepartment.supervisor == g.currentUser.supervisor)
+        allowedDepartmentIds = [sd.department_id for sd in supervisorDepts]
         if dept.departmentID not in allowedDepartmentIds:
             return render_template('errors/403.html'), 403
 
@@ -96,6 +98,7 @@ def managePositions(org, account):
                            department_name = dept.DEPT_NAME,
                            positions = positions
                            )
+
 @main_bp.route('/supervisorPortal/download', methods=['POST'])
 def downloadSupervisorPortalResults():
     '''
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -59,25 +59,23 @@ def departmentPortal(org=None,account=None):
     except (NameError, DoesNotExist):
         dept = None
 
-
     if g.currentUser.isLaborAdmin:
         departments = list(Department.select().order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
     else:
         departments = list(getDepartmentsForSupervisor(g.currentUser).order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
+
+    positionsList, posURL = getActivePositions(dept)
 
     if dept and dept.departmentID not in [d.departmentID for d in departments]:
         return render_template('errors/403.html'), 403
 
     return render_template('main/departmentPortal.html', 
                            departments = departments,
-                        department = dept,
-                        )
-
+                           department = dept,
+                           positions = positionsList,
+                           posURL = posURL)
 @main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
 def managePositions(org, account):
-    if not g.currentUser or not g.currentUser.supervisor: 
-        return render_template('errors/403.html'), 403
-
     try:
         dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
     except DoesNotExist:
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -12,7 +12,7 @@ from app.models.term import Term
 from app.controllers.admin_routes.allPendingForms import checkAdjustment
 from app.controllers.main_routes import main_bp
 from app.logic.download import CSVMaker, saveFormSearchResult, retrieveFormSearchResult
-from app.logic.search import getDepartmentsForSupervisor, isDepartmentAllowed, searchPerson, searchSupervisorPortal
+from app.logic.search import getDepartmentsForSupervisor, searchPerson, searchSupervisorPortal
 from app.login_manager import require_login, logout
 from app.logic.getTableData import getDatatableData
 from app.logic.banner import Banner
@@ -82,8 +82,10 @@ def managePositions(org, account):
     except DoesNotExist:
         return render_template('errors/404.html'), 404
 
-    if not isDepartmentAllowed(g.currentUser, dept):
-        return render_template('errors/403.html'), 403
+    if not g.currentUser.isLaborAdmin:
+        allowedDepartmentIds = [d.departmentID for d in getDepartmentsForSupervisor(g.currentUser)]
+        if dept.departmentID not in allowedDepartmentIds:
+            return render_template('errors/403.html'), 403
 
     positions = (PositionHistory.select()
                                 .where((PositionHistory.department == dept) &
--- a/app/controllers/main_routes/main_routes.py
+++ b/app/controllers/main_routes/main_routes.py
@@ -67,11 +67,10 @@ def departmentPortal(org=None,account=None):
     if dept and not g.currentUser.isLaborAdmin and dept.departmentID not in [d.departmentID for d in departments]:
         return render_template('errors/403.html'), 403
 
-    positionsList, posURL = getActivePositions(dept)
-
     return render_template('main/departmentPortal.html', 
                            departments = departments,
-                           department = dept)
+                        department = dept,
+                        )
 
 @main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
 def managePositions(org, account):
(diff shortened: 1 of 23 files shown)


app/controllers/main_routes/main_routes.py
```diff
@@ -63,26 +64,40 @@ def departmentPortal(org=None,account=None):
     else:
         departments = list(getDepartmentsForSupervisor(g.currentUser).order_by(Department.isActive.desc(), Department.DEPT_NAME.asc()))
 
+    if dept and not g.currentUser.isLaborAdmin and dept.departmentID not in [d.departmentID for d in departments]:
```
BrianRamsay: We can simplify this by removing the check for labor admin because we are already accounting for labor admins when we choose departments above this.

app/controllers/main_routes/main_routes.py
```diff
@@ -63,26 +64,40 @@ def departmentPortal(org=None,account=None):
...
-            return "True"
-    
-    except Exception as e:
-        print(f'Could not add user to department: {e}')
-        return "", 500
+@main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
+def managePositions(org, account):
+    currentUser = require_login()
+    if not currentUser or not currentUser.supervisor:
+        return render_template('errors/403.html'), 403
```
BrianRamsay: This check is simpler if we use `g.currentUser`, which is the new way and present in all requests (or else they would have been redirected to login).

app/controllers/main_routes/main_routes.py
```diff
@@ -63,26 +64,40 @@ def departmentPortal(org=None,account=None):
...
 
+    try:
+        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+    except DoesNotExist:
+        return render_template('errors/404.html'), 404
+
+    if not currentUser.isLaborAdmin:
+        allowedDepartmentIds = [d.departmentID for d in getDepartmentsForSupervisor(currentUser)]
+        if dept.departmentID not in allowedDepartmentIds:
```
BrianRamsay: You should be able to check dept in the returned list directly, without needing an intermediate list, like `if dept in getDepartmentsForSupervisor...`. Though I'm not 100% sure, as the function returns a modelselect, I think. Really, we do this check enough across all of the new functionality we should probably have an `isDepartmentAllowed` boolean method that takes a supervisor and department and does this logic internally

app/static/css/managePositions.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.department-header{
+    padding:1%;
+    margin-top:-20px;
+}
+
+td{
+    text-align:center !important;
+}/* managePositions.css */
```
BrianRamsay: There's got to be a better way to center text in a cell than doing a blanket `!important` on every table cell

app/static/css/managePositions.css
```diff
@@ -0,0 +1,33 @@
...
+}
+.margin-top-10{
+    margin-top:-25px;
+}
+*{
+    /* outline:solid 1px lime; */
+    margin:0;
+    padding:0;
+    box-sizing:border-box;
+}
```
BrianRamsay: do we really need to reset every element?

app/static/css/managePositions.css
```diff
@@ -0,0 +1,33 @@
+.width-12{
+    width:12%;
+}
+.margin-top-10{
+    margin-top:-25px;
+}
```
BrianRamsay: Why does the name not match the rule?

app/static/css/managePositions.css
```diff
@@ -0,0 +1,33 @@
+.width-12{
+    width:12%;
+}
```
BrianRamsay: You only use this on <th> in this page, so you could just make <th> have this width.

BrianRamsay (changes requested):

app/controllers/main_routes/main_routes.py
conwelld: There seems to be a lot of nuance for this implementation. From what I see, the cookie will "hold" the status, indicating that a user is a labor admin. This can be changed using an incognito window and by changing the debugging user from heggens to pearcej in the secret config. However, this only lets you test the CS department, which could be throwing the 403 error we want, since she does not have access to other departments anyway. I feel the previous implementation we have here works better because it leads down a rabbit hole

app/controllers/main_routes/main_routes.py
```diff
@@ -59,36 +59,43 @@ def departmentPortal(org=None,account=None):
...
                            department = dept,
                            positions = positionsList,
                            posURL = posURL)
-
-@main_bp.route('/supervisorPortal/addUserToDept', methods=['GET', 'POST'])
-def addUserToDept():
-    userDeptData = request.form
-    supervisorDeptRecord = SupervisorDepartment.get_or_none(supervisor = userDeptData['supervisorID'], department = userDeptData['departmentID'])
+@main_bp.route('/department/<org>/<account>/positions', methods=['GET'])
+def managePositions(org, account):
```
MImran2002: there is a newly created departmentPortal.py controller I think anything route that doesn't stays on the three card department page should be in the departmentPortal.py but not mainroute. @BrianRamsay please correct me if I am wrong.

app/controllers/main_routes/main_routes.py
```diff
@@ -59,36 +59,43 @@ def departmentPortal(org=None,account=None):
...
 
-        else:
-            SupervisorDepartment.create(supervisor=userDeptData['supervisorID'], department=userDeptData['departmentID'])
-            return "True"
-    
-    except Exception as e:
-        print(f'Could not add user to department: {e}')
-        return "", 500
+    if not g.currentUser.isLaborAdmin:
+        allowedDepartmentIds = [d.departmentID for d in getDepartmentsForSupervisor(g.currentUser)]
```
MImran2002: getDepartmentForSupervisor function returns all the departments the user is asscoiated meaning even if they don't exist in the department based on supervisordepartment table they can still see those departments because somewhere they have a lsf assigned to them a form history assigned to them. So we don't want to use that.

app/controllers/main_routes/main_routes.py
```diff
@@ -59,36 +59,43 @@ def departmentPortal(org=None,account=None):
...
+        if dept.departmentID not in allowedDepartmentIds:
+            return render_template('errors/403.html'), 403
 
+    positions = (PositionHistory.select()
```
MImran2002: getActivePositions in getpositions.py logic is a great place to either add your logic or refactor the existing getactivepositionlogic to return all position or active position base on parameters.

MImran2002 (changes requested):

app/controllers/main_routes/main_routes.py
conwelld: through supervisor department, not just the ones they have a form history for,

app/controllers/main_routes/main_routes.py
conwelld: Refactor the department check; use multiple functions with parameters for a query, or an if statement if the query returns anything. Within the getDepartments function.

app/controllers/main_routes/main_routes.py
```diff
@@ -64,7 +64,7 @@ def departmentPortal(org=None,account=None):
     except (NameError, DoesNotExist):
         dept = None
 
-    if currentUser.isLaborAdmin:
+    if g.currentUser.isLaborAdmin:
```
MImran2002: you can use currentUser here why did you use g.currentUser instead.

app/controllers/main_routes/main_routes.py
```diff
@@ -73,6 +73,9 @@ def departmentPortal(org=None,account=None):
 
     positionsList, posURL = getActivePositions(dept) 
 
+    if dept and dept.departmentID not in [d.departmentID for d in departments]:
```
MImran2002: why is this added?

app/logic/getPositions.py
```diff
@@ -18,4 +18,10 @@ def getActivePositions(dept):
             positionsList.append(i.positionTitle + ": " + "(WLS " + str(i.wls) + ")")
             posURL.append(str(i.positionCode))
 
-    return positionsList, posURL
\ No newline at end of file
+    return positionsList, posURL
+
+def supervisorsPositions(dept):
```
MImran2002: this function will be more reasonable if tis named getPositions.

MImran2002 (changes requested):
just these and I will be able to merge it

app/controllers/main_routes/main_routes.py
```diff
@@ -73,6 +73,9 @@ def departmentPortal(org=None,account=None):
 
     positionsList, posURL = getActivePositions(dept) 
 
+    # if dept and dept.departmentID not in [d.departmentID for d in departments]:
```
MImran2002: remove this comment

app/templates/main/managePositions.html
```diff
@@ -0,0 +1,45 @@
...
+    <h1 class="text-center">{{ department_name }} Positions</h1>
+
+    <div class="table-responsive">
+        <table id="positionTable" class="table table-striped table-bordered text-center">
+            <thead>
+                <tr>
+                    <th class="text-center width-46">Position</th>
+                    <th class="text-center width-12">WLS</th>
+                    <th class="text-center width-30">Last Revision</th>
+                    <th class="text-center width-33">View / Edit</th>
```
MImran2002: Action not view/edit

app/templates/main/managePositions.html
```diff
@@ -0,0 +1,45 @@
...
+            </thead>
+            <tbody>
+                {% for position in positions %}
+                <tr>
+                    <td>{{ position.positionTitle }} <small class="text-muted">{{ position.positionCode }}</small></td>
+                    <td>{{ position.wls }}</td>
+                    <td>{{ position.revisionDate }}</td>
+                    <td>
+                        <button class="btn btn-success view-btn" data-position-id="{{ position.id }}">View</button>
+                        <button class="btn btn-primary request-btn" data-position-id="{{ position.id }}">Edit Position</button>
```
MImran2002: it is revise position not edit position

MImran2002 (changes requested):

MImran2002 (approved):

PR #662, opened 2026-08-03
Annual Position Review Button

johnlolonga19:
Fixes issue [#618](https://github.com/BCStudentSoftwareDevTeam/lsf/issues/618) 

- The Labor Office should be able to initiate the annual position review process for every department from the Manage Departments page. 

- This workflow requests departments to review their position descriptions and submit any necessary updates before the next academic year.

## Additions

image

image

- A working "Annual Position Review" button on the Manage Departments page. Clicking it opens a confirmation window explaining what's about to happen.

- An "Edit Email Template" option in that confirmation window, so an admin can review or change the wording of the email before sending it.

- A "Submit Request" button that actually sends the request. Once sent, every active department's supervisors and Labor Coordinators receive an email asking them to review their positions for the current academic year.

- A confirmation message on screen after sending, showing how many departments the request went out to.
The system now keeps a record of when a review was requested for each department and academic year, and who requested it, so re-sending a request updates that record instead of creating duplicates.

## Changes

- `annualPositionReview.html` — the previously-stubbed confirmation modal (empty onclick="" on every button) is now fully wired: Submit Request calls the new JS, Edit Email Template links to `emailTemplates` linked to this template

- `manageDepartments.html`  loads `annualPositionReview.js`; the trigger button now carries data-academic-year so the JS knows which academic year is selected.

- manageDepartments() view now also passes the chosenAY Term object into the template context (previously only chosenAY.termName was passed), needed for the data-academic-year attribute above.

- `emailTemplates.js`  added prefillFromQueryParams() so the Email Templates editor can be linked with Recipient/Form Type/Action pre-selected (used by the Edit Email Template link above)

- The confirmation page and its buttons (Cancel, Edit Email Template, Submit Request) previously did nothing when clicked they're now fully functional.

- The email template editor page can now be opened directly from the confirmation window with the right template already selected, instead of requiring the admin to fill it manually.

- Added the test suite for `annualPositionReview.py`

- Moved the logic from `annualPositionReview.py` to `emailHandler.py` to avoid duplicating the function

## Testing

- ALWAYS_SEND_MAIL was set to False in local secret_config.yaml (untracked, not part of this PR) so local testing doesn't attempt a real SMTP send.

- To submit an annual position review, on the sidebar menu, click Admin
- Then click the " Manage Department " that will take you to the page 
- On the page click "Annual Position Review" and then " Submit Request"
- A message will show that the request has been sent to x departments out of y. 

- Two ways to edit email template through Manage Department and Manage Email Template:
     - To edit the email template inside the, click on "Edit Email Template" that will prefill the Position Review components 
     - Or inside Manage Email Template, fill the fields for the Position Review Email and will allow you to edit
     - And you can save the changes or discard


app/logic/annualPositionReview.py
```diff
@@ -0,0 +1,75 @@
...
+    term = Term.get(Term.termCode == academicYearTermCode)
+    template = EmailTemplate.get(EmailTemplate.purpose == "Annual Position Review Request")
+    departments = Department.select().where(Department.isActive == True)
+
+    sentCount = 0
+    for department in departments:
+        supervisors, laborCoordinators = getSupervisors(department)
+        recipients = {person["email"] for person in supervisors + laborCoordinators if person["email"]}
+        if not recipients:
+            continue
```
BhushanSah: Should we still create or update the PositionReview record for this department even if there are no email recipients?

The PR description says the system keeps a record of when a review was requested for each department and academic year. Right now, departments with no supervisors/coordinators are skipped before the PositionReview record is created, so those departments will have no record of the request.

app/static/js/annualPositionReview.js
```diff
@@ -0,0 +1,33 @@
...
+/*
+ POSTs the Annual Position Review request for the currently selected academic year.
+ Sends a review request email to every active department's Labor Coordinators and
+ supervisors, then shows a success/failure flash message.
+
+ RETURNS: None
+*/
+  var academicYear = $('[data-target="#annualPositionModal"]').data('academic-year');
+
+  $.ajax({
```
BhushanSah, line 17: Can we add an error callback here?

Right now, if the server returns 403, 500, or the request fails, the modal will likely stay open and the user will not get a clear message. The success handler handles {"Success": false}, but it does not handle actual AJAX errors.

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -93,6 +95,27 @@ def complianceStatusCheck():
...
+    Labor Coordinators and supervisors for the selected academic year, and
+    records the request. Triggered from the Manage Departments page.
+    """
+    currentUser = require_login()
+    if not currentUser or not currentUser.isLaborAdmin:
+        return jsonify({"Success": False}), 403
+
+    try:
+        rsp = request.get_json()
+        result = sendAnnualPositionReviewRequests(int(rsp['academicYear']), currentUser)
```
BhushanSah: Can we validate `rsp` and `academicYear` before calling int()?

If the request body is missing, not JSON, or does not include academicYear, this will raise an exception and return {"Success": false} with a 200 response. It would be cleaner to return a 400 with a clear message for bad input.

app/logic/annualPositionReview.py
```diff
@@ -0,0 +1,75 @@
...
+
+from app import app
+from app.models.department import Department
+from app.models.emailTemplate import EmailTemplate
+from app.models.positionReview import PositionReview
+from app.models.term import Term
+from app.logic.getSupervisors import getSupervisors
+
+
+def sendMail(mail, message: Message):
```
BhushanSah: This sendMail function looks very similar to the existing emailHandler.send logic. Can we reuse the existing email sending helper or extract the shared behavior so we do not have two versions of the same mail override / reply_to / testing behavior?

tests/code/test_annualPositionReview.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+
+@pytest.mark.integration
+def test_sendAnnualPositionReviewRequests():
+    with app.app_context():
+        with mainDB.atomic() as transaction:
+            Department.update(isActive=False).where(Department.isActive == True).execute()
+
+
```
BhushanSah, line 24: This logic test is helpful. Can we also add a route-level test for `/admin/manageDepartments/annualPositionReview`?

I think we should test that non-admin users get 403 and that missing or invalid academicYear input returns a clear error. That would cover the endpoint behavior, not just the helper function.

app/templates/snips/annualPositionReview.html
```diff
@@ -5,17 +5,17 @@
...
             <div class="modal-body" id="annualPositionModalBody">
                 <p align="left">Submitting this request will send an email to all departments. The email will ask them to review their positions for the current academic year.</p>
             </div>
             <div class="modal-footer">
-                <button type="button" class="btn btn-secondary" data-dismiss="modal" onclick="">Cancel</button>
-                <button type="button" class="btn btn-info" onclick="">Edit Email Template</button>
-                <button type="button" class="btn btn-primary" onclick="">Submit Request</button>
+                <button type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
+                <a class="btn btn-info" href="/admin/emailTemplates?audience=Department&formType=Position%20Review&action=Annual%20Request">Edit Email Template</a>
+                <button type="button" class="btn btn-primary" onclick="submitAnnualPositionReview()">Submit Request</button>
```
BhushanSah: Can we disable this Submit Request button while the AJAX request is running?

Since this sends emails to all active departments, a double-click could send the same email request more than once. The PositionReview table avoids duplicate records, but the emails could still go out multiple times.

BhushanSah (changes requested):

app/logic/annualPositionReview.py
```diff
@@ -0,0 +1,75 @@
...
+    mail = Mail(app)
+    term = Term.get(Term.termCode == academicYearTermCode)
+    template = EmailTemplate.get(EmailTemplate.purpose == "Annual Position Review Request")
+    departments = Department.select().where(Department.isActive == True)
+
+    sentCount = 0
+    for department in departments:
+        supervisors, laborCoordinators = getSupervisors(department)
+        recipients = {person["email"] for person in supervisors + laborCoordinators if person["email"]}
+        if not recipients:
```
NYABUTOA: when a department has no supervisors/coordinators with an email, it's skipped with zero logging. No way to tell from logs which departments were skipped or why.

app/static/js/annualPositionReview.js
```diff
@@ -0,0 +1,33 @@
...
+      $("#annualPositionModal").modal("hide");
+
+      var category, msg;
+      if (response["Success"]) {
+        category = "success";
+        msg = "Position review requests sent to " + response["sentCount"] + " of " + response["departmentCount"] + " departments.";
+      } else {
+        category = "danger";
+        msg = "Something went wrong sending the Annual Position Review requests.";
+      }
```
NYABUTOA: the UI only shows "Success" or "Something went wrong," so an admin can't tell which departments were reached before a failure.

app/logic/annualPositionReview.py
```diff
@@ -0,0 +1,75 @@
...
+            PositionReview.create(
+                academicYear=term,
+                department=department,
+                requestedOn=datetime.now(),
+                requestedBy=requestingUser
+            )
+
+        sentCount += 1
+        print("Sent Annual Position Review request to {} for department {}.".format(", ".join(recipients), department.DEPT_NAME))
+        print("{} Annual Position Review requests sent for academic year {}.".format(sentCount, term.termName))
```
ArtemKurasov: Were lines 72 and 73 used for debugging? If so, then **remove them**

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -93,6 +95,27 @@ def complianceStatusCheck():
...
+        return jsonify({"Success": True, **result})
+    except Exception as e:
+        print(e)
```
ArtemKurasov: Do we need this print statement (line 114)?

ArtemKurasov (changes requested):
The code is good overall, but I saw several print statements in the code. You should consider deleting them if they serve no purpose other than debugging

app/logic/annualPositionReview.py
johnlolonga19: i just got rid of `annualPositionReview.py` and move both the `sendAnnualPositionReviewRequests()` and `sendMail()` into emailHandler.py so that i can just use the function as send() to remove the duplicate

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -93,6 +95,34 @@ def complianceStatusCheck():
 
 
 
+@admin.route('/admin/manageDepartments/annualPositionReview', methods=['POST'])
+def annualPositionReviewRequest():
+    """
+    Sends an Annual Position Review request email to every active department's
```
MImran2002, line 102: using g.openterm will create a lot of bugs in case user have two open term, if the position review is for next ay while the openterm is for this ay

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -93,6 +95,34 @@ def complianceStatusCheck():
...
+    Labor Coordinators and supervisors for the selected academic year, and
+    records the request. Triggered from the Manage Departments page.
+    """
+    currentUser = require_login()
+    if not currentUser or not currentUser.isLaborAdmin:
```
MImran2002: we also need to give access to isLaborDepartmentStudent as well I have checked the controllers and we do give access.

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -93,6 +95,34 @@ def complianceStatusCheck():
...
+        return jsonify({"Success": False}), 403
+
+    rsp = request.get_json(silent=True)
+    if not rsp or "academicYear" not in rsp:
```
MImran2002: from line 110 to 122 can be consolidate into two try and except with the first try doing academic year and second try doing result and returning success jsonify.

app/logic/annualPositionReview.py
```diff
@@ -0,0 +1,78 @@
...
+from app import app
+from app.models.department import Department
+from app.models.emailTemplate import EmailTemplate
+from app.models.positionReview import PositionReview
+from app.models.term import Term
+from app.logic.getSupervisors import getSupervisors
+
+
+def sendMail(mail, message: Message):
+    if app.config['ENV'] == 'production' or app.config['ALWAYS_SEND_MAIL']:
```
MImran2002: emailHandler page has a emailhandler class but it runs on id, that means you will have to expand the emailhandler to be flexible. This means emailhandler can have a second optional parameter and turn the id parameter into a default too. For the second parameter turn it to default and have a condition in the init so that when one of the parameter is added the class knows what to create. 

This means you won't need this file while you incorporate sendannualpositionreviewrequests into the emailhandler function which is good as we would be doing similar with allocation request

app/models/positionReview.py
```diff
@@ -0,0 +1,14 @@
+from app.models import *
```
MImran2002, line 1: we should try to work off from positionhistory model

app/static/js/annualPositionReview.js
```diff
@@ -0,0 +1,50 @@
...
+        : "Something went wrong sending the Annual Position Review requests.";
+      flashMessage("danger", msg);
+    },
+    complete: function() {
+      $(button).prop("disabled", false);
+    }
+  })
+}
+
+function flashMessage(category, msg) {
```
MImran2002, line 47: there is a flash message function in base.js

database/migrate_db.sh
```diff
@@ -32,6 +32,7 @@ pem add app.models.studentLaborEvaluation.StudentLaborEvaluation
 pem add app.models.formSearchResult.FormSearchResult
 pem add app.models.positionHistory.PositionHistory
 pem add app.models.allocation.Allocation
+pem add app.models.positionReview.PositionReview
```
MImran2002, line 35: we can remove this

MImran2002 (changes requested):

app/static/js/annualPositionReview.js
GalinaP7: Lines 33-422 do AJAX error callback. Lines 38-40 for 403 & fallback for 500/network errors.

tests/code/test_annualPositionReview.py
GalinaP7: Added route-level tests covering unauthorized users receiving 403, missing academicYear returning 400, and invalid academicYear returning 400 with a clear error message. The three new tests pass independently.

app/logic/annualPositionReview.py
GalinaP7: Departments without valid email recipients are now logged with the department name and reason before email delivery is skipped. The PositionReview record is still created or updated.

app/static/js/annualPositionReview.js
GalinaP7: Added per-department failure handling. The UI now reports how many departments were reached and lists any departments whose emails failed.

app/controllers/admin_routes/manageDepartments.py
GalinaP7: This is fixed. The review uses the selected academic year from the request, not g.openTerm, so multiple open terms won’t cause issues.

app/controllers/admin_routes/manageDepartments.py
GalinaP7: Access is allowed for both labor admins and labor department students. Already fixed.

app/models/positionReview.py
GalinaP7: *remove positionReview and use positionHistory to replace it (EVERYWHERE)

MImran2002:
There is a merge conflict happening

app/logic/manageDepartments.py
```diff
@@ -14,6 +14,34 @@
 
 
 
+def generateAdjacentYears(academicYearTermCode=None):
+    """
+    Generates the current and the following academic years.
+    """
+
+    currentYear      = g.currentAY[0]
+    nextYear         = g.currentAY[1]
```
MImran2002: can be written as `currentYear, nextYear = g.currentAY` because g.currentAY for example returns a tuple of (2026, 2027)

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -38,8 +40,16 @@ def manageDepartments():
         elif currentUser.supervisor:
             return render_template('errors/403.html'), 403
 
-    currentAY, nextAY = getCurrentAndNextAY()
-    chosenAY = Term.get(Term.termCode == currentAY.termCode)
+
+    # The condition below may be deleted if the routing to the Manage Departments page is changed.
+    if academicYear == None:
```
MImran2002: `academicYear = g.currentAY[0] * 100 if academicYear is None else int(academicYear)`

MImran2002: This if else can be consolidated in a conditional expression

database/demo_data.py
```diff
@@ -709,27 +709,18 @@
         "termStart": "2026-08-01",
         "termEnd": "2027-05-01",
         "termState": 0,
-        "primaryCutOff": "2026-09-01",
-        "adjustmentCutOff": "2026-10-01",
-    },
-    {
-        "termCode": "202601",
```
MImran2002: removing this breaks the demo test data being created when reset_database.sh is run

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -85,6 +96,36 @@ def complianceStatusCheck():
...
+@admin.route('/admin/manageDepartments/annualPositionReview', methods=['POST'])
+def annualPositionReviewRequest():
+    """
+    Sends an Annual Position Review request email to every active department's
+    Labor Coordinators and supervisors for the selected academic year, and
+    records the request. Triggered from the Manage Departments page.
+    """
+    currentUser = require_login()
+    if not currentUser or not (currentUser.isLaborAdmin or currentUser.isLaborDepartmentStudent):
+        return jsonify({"Success": False}), 403
```
MImran2002: instead of jsonify({"success":false}) we usually use render_template('errors/403.html')

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -85,6 +96,36 @@ def complianceStatusCheck():
...
+
+    rsp = request.get_json(silent=True)
```
MImran2002: this rsp is receiving the chosen academic year from javascript from the front-end we no longer allow the users to chose a academic year.

MImran2002: to select a academic year we are doing position review for the positions in each department means this likely is a interval of either 3 years. we are not allowing the labor department to select which year its more of in intervals the positions are being reviewed.

app/controllers/admin_routes/manageDepartments.py
```diff
@@ -85,6 +96,36 @@ def complianceStatusCheck():
...
+
+    try:
+        academicYear = int(rsp["academicYear"])
+    except (TypeError, ValueError, KeyError):
+        return jsonify({"Success": False, "message": "Request must include a valid academicYear."}), 400
+
+    try:
+        handler = emailHandler(academicYearTermCode=academicYear)
```
MImran2002: not handler the variable would most likely be email something like positionReviewEmail.

app/logic/emailHandler.py
```diff
@@ -16,82 +16,96 @@
...
 from datetime import datetime, date
+from app.models.department import Department
+from app.models.term import Term
+from app.models.positionHistory import PositionHistory
+from app.logic.getSupervisors import getSupervisors
 
 
 class emailHandler():
-    def __init__(self, formHistoryKey):
+    def __init__(self, formHistoryKey=None, academicYearTermCode=None):
```
MImran2002: academicYearTermCode is general like how formHistoryKey is not related to email but it is a distinction of what the emailhandler do we should turn the parameter naming of academicYearTermCode to something that resembles its purpose. This academicYearTermCode is going to use in two instance the allocation review email and position review email. So the parameter naming can be termCodeUnderReview, targetTermCode and more

app/logic/emailHandler.py
```diff
@@ -110,7 +124,67 @@ def send(self, message: Message):
...
+        """
+        template = EmailTemplate.get_or_none(EmailTemplate.purpose == "Annual Position Review Request")
+        if template is None:
+            raise ValueError("The 'Annual Position Review Request' email template is missing.")
+
+        departments = Department.select().where(Department.isActive == True)
+
+        sentCount = 0
+        failedDepartments = []
+        for department in departments:
```
MImran2002: this logic is a function on its own emailHandler methods are only here to handle different email logic.

MImran2002: existingReview = PositionHistory.get_or_none(
                PositionHistory.academicYear == self.term,
                PositionHistory.department == department,
                PositionHistory.positionCode.is_null(True)
            )
            if existingReview:
                existingReview.requestedOn = datetime.now()
                existingReview.requestedBy = requestingUser
                existingReview.save()
            else:
                PositionHistory.create(
                    academicYear=self.term,
                    department=department,
                    requestedOn=datetime.now(),
                    requestedBy=requestingUser
                )

app/logic/emailHandler.py
```diff
@@ -110,7 +124,67 @@ def send(self, message: Message):
...
+            # A review is considered "requested" for every active department as soon
+            # as this runs, whether or not there's currently anyone to email - a
+            # department with no supervisors/coordinators assigned is itself worth
+            # surfacing, not silently skipping.
+            existingReview = PositionHistory.get_or_none(
+                PositionHistory.academicYear == self.term,
+                PositionHistory.department == department,
+                PositionHistory.positionCode.is_null(True)
```
MImran2002: this logic feels iffy. the idea that we are considering existingreview to be based off from positioncode being null isn't a good logic here. I would encourage to see the composite key in positionhistory and see whether a position should be reviewed based on how long it has not been reviewed and what status it is. i believe status wise we should be reviewing the active status.

app/logic/emailHandler.py
```diff
@@ -110,7 +124,67 @@ def send(self, message: Message):
...
+            )
+            if existingReview:
+                existingReview.requestedOn = datetime.now()
+                existingReview.requestedBy = requestingUser
+                existingReview.save()
+            else:
```
MImran2002: I don't think reviewing means creating a positionhistory

app/models/positionReview.py
MImran2002: we don't need a positionreview table. we can base off the logic and insert data with positionhistory

app/static/js/annualPositionReview.js
```diff
@@ -0,0 +1,47 @@
+function submitAnnualPositionReview(button) {
+/*
+ POSTs the Annual Position Review request for the currently selected academic year.
+ Sends a review request email to every active department's Labor Coordinators and
+ supervisors, then shows a success/failure flash message.
+
+ RETURNS: None
+*/
+  var academicYear = $('[data-target="#annualPositionModal"]').data('academic-year');
```
MImran2002: as we are not selecting academic year this is not needed. Like if you go check out the manage departments page you will see there is no longer an option to select a academic year.

app/static/js/annualPositionReview.js
```diff
@@ -0,0 +1,47 @@
...
+  // more than once. Re-enabled in complete regardless of outcome so a retry
+  // after a failure is possible without reloading the page.
+  $(button).prop("disabled", true);
+
+  $.ajax({
+    method: "POST",
+    url: "/admin/manageDepartments/annualPositionReview",
+    dataType: "json",
+    contentType: "application/json",
+    data: JSON.stringify({"academicYear": academicYear}),
```
MImran2002: this is not needed here as the backend doesn't need the AY term code from front-end to have its logic works.

app/static/js/emailTemplates.js
```diff
@@ -13,10 +13,61 @@ function getEmailArray() {
...
+  if (!recipient) {
+    return;
+  }
+
+  // populatePurpose() ends by inserting the body into the CKEditor instance,
+  // which initializes asynchronously (CKEDITOR.replace() in emailTemplates.html).
+  // Running this cascade immediately on page load (unlike a human clicking
+  // through the dropdowns) can race ahead of that, so wait for the editor
+  // to be ready before touching it.
+  function runPrefill() {
```
MImran2002: run prefill function should be outside of this function because it restricts the scope of how the function can be used.

MImran2002 (changes requested):

PR #664, opened 2026-08-04
Edit position page

NYABUTOA:
Issue Description - 

Added a Revise Position page for admins to edit a position's details and description.

Changes -

Added revisepositionpage.html, the new revise page. It's a form pre-filled with the position's title, WLS level, last revision date, revised-by, and all of its existing description sections so admins aren't starting from a blank page.
Added revisepositionpage.js so admins can add or remove description sections 
Added revisepositionpage.css for the section-row and action-button styling the form needed that Bootstrap didn't already cover.
Added a link on the "Revise Position" button in managePositions.html so it takes you to the revise page for that exact position.

Screenshot 2026-08-04 at 4 48 05 PM

Test -

Went to Manage Positions and clicked "Revise Position" on a row, confirmed it opens the revise page and pulls in that position's title, code, WLS level, last revision date, revised-by, and its existing sections correctly.
Clicked "Add Section" a few times and confirmed each click adds a fresh, empty section row.
Clicked "Remove Section" and confirmed it only removes that one row, not the others.
Clicked "Cancel" and confirmed it takes you back to Manage Positions for the right department.
Reloaded the page to make sure the pre-filled data still shows up the same way.


app/templates/main/managePositions.html
```diff
@@ -33,8 +33,8 @@ <h1 class="text-center">{{ department_name }} Positions</h1>
                     <td>{{ position.wls }}</td>
                     <td>{{ position.revisionDate }}</td>
                     <td>
-                        <button class="btn btn-success view-btn" data-position-id="{{ position.id }}">View</button>
-                        <button class="btn btn-primary request-btn" data-position-id="{{ position.id }}">Revise Position</button>
+                        <a href="{{ url_for('main.postionDescription', org=department.ORG, account=department.ACCOUNT, positionCode=position.positionCode) }}" class="btn btn-success">View</a>
+                        <a href="{{ url_for('main.revisePosition', org=department.ORG, account=department.ACCOUNT, positionCode=position.positionCode) }}" class="btn btn-primary">Revise Position</a>
```
BhushanSah, line 37: This currently points to `main.revisePosition`, but I do not see a `revisePosition` route added in this PR. Because of that, rendering Manage Positions may fail with a BuildError before the user can even click the button.

Can you add the route/controller logic for `revisePosition`, or update this link to point to the correct existing endpoint?

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,10 +1,72 @@
...
+
+from app.controllers.main_routes import main_bp
+from app.logic.download import makePositionDescriptionPDF
+from app.logic.getPositions import getPosition, getPositions, getPositionDescriptionSections
 from app.models.department import Department
+from app.models.positionHistory import PositionHistory
 from app.models.supervisorDepartment import SupervisorDepartment
 
+@main_bp.route('/department/<org>/<account>/positions/<positionCode>', methods=['GET'])
+def postionDescription(org, account, positionCode):
```
BhushanSah, line 14: These new position view/download routes should use the same authorization check as Manage Positions. Right now a user who knows the department org/account and position code can directly view or download the position description without being a labor admin or a supervisor assigned to that department.

Can you add the same labor admin / SupervisorDepartment check here before returning the page or PDF?

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,10 +1,72 @@
...
+        'main/individualPositions.html',
+        department=dept,
+        position=position,
+        sections=sections
+    )
+
+
+@main_bp.route('/department/<org>/<account>/positions/<positionCode>/download', methods=['GET'])
+def downloadPositionDescription(org, account, positionCode):
+    try:
```
BhushanSah, line 45: These new position view/download routes should use the same authorization check as Manage Positions. Right now a user who knows the department org/account and position code can directly view or download the position description without being a labor admin or a supervisor assigned to that department.

Can you add the same labor admin / SupervisorDepartment check here before returning the page or PDF?

app/controllers/main_routes/departmentPortal.py
BhushanSah, line 14: Small naming issue: `postionDescription` looks like a typo. It should probably be `positionDescription`.

Also, if templates use `url_for`, the endpoint name will change with the function name, so the template link should be updated too.

tests/code/test_getPositions.py
```diff
@@ -80,7 +80,62 @@ def test_getActivePositions():
         assert len(posURL3) == 0
 
         transaction.rollback()
-    
+
+@pytest.mark.integration
```
BhushanSah, line 84: These tests still create `PositionHistory` records with `description=""`, but the model no longer has a `description` field and now requires `revisedBy`.

Can you update the tests to use `revisedBy` instead? Several creates in this file may fail because `revisedBy` is required and not provided.

app/controllers/main_routes/main_routes.py
```diff
@@ -15,7 +15,7 @@
 from app.controllers.admin_routes.allPendingForms import checkAdjustment
 from app.controllers.main_routes import main_bp
 
-from app.logic.download import CSVMaker, saveFormSearchResult, retrieveFormSearchResult
+from app.logic.download import CSVMaker, saveFormSearchResult, retrieveFormSearchResult, makePositionDescriptionPDF
```
BhushanSah, line 18: `makePositionDescriptionPDF` is imported here, but I do not see it being used in this file. Since `download.py` already imports from main_routes, adding this import here may also make the circular dependency harder to follow.

Can this import be removed from `main_routes.py` and kept only where it is actually used?

app/logic/getPositions.py
```diff
@@ -20,8 +21,38 @@ def getActivePositions(dept):
 
     return positionsList, posURL
 
+def getPosition(dept, positionCode, revisionDate=None):
+    """
+    Returns a single position for a given department, position code, and optional revision date. 
+    If no revision date is provided, the most recent revision is returned.
+    """
+    positionQuery = PositionHistory.select().where(
```
BhushanSah, line 31: For the short Peewee queries here, can we keep them in one line or clean up the formatting? For example, this can be simpler:

`return PositionHistory.select().where(PositionHistory.department == dept, PositionHistory.status == "Active").order_by(PositionHistory.positionTitle.asc())`

The current `return((...` formatting is harder to read and does not match the style used elsewhere.

BhushanSah (changes requested):

app/controllers/main_routes/departmentPortal.py
NYABUTOA: Andrew is working on this

NYABUTOA: Andrew is working on this

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,10 +1,105 @@
...
+        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+    except (NameError, DoesNotExist):
+        return render_template('errors/404.html'), 404
+
+    revisionDateParam = request.args.get('revisionDate')
+    revisionDate = None
+    if revisionDateParam:
+        try:
+            revisionDate = datetime.strptime(revisionDateParam, '%Y-%m-%d').date()
```
brightfietsop-ux, line 24: Currently, when a section is added and saved, it works. But if the page is reloaded to check if it saves in the database, it breaks with an unexpected error.  I believe it might be a routing issue or simply simply just an exception handler problem

app/controllers/main_routes/departmentPortal.py
```diff
@@ -1,10 +1,105 @@
...
+
+    sections = getPositionDescriptionSections(position)
+
+    return render_template(
+        'main/individualPositions.html',
+        department=dept,
+        position=position,
+        sections=sections
+    )
+
```
brightfietsop-ux, line 41: make sure the WLS number entered should not be more than 6. until it gets revised to the new 1-3 system. currently, any number input is accepted, even if the number is a 10000 etc.

brightfietsop-ux (changes requested):

app/templates/main/revisepositionpage.html
```diff
@@ -0,0 +1,105 @@
+{% extends "base.html" %}
+
```
MImran2002, line 2: when I look into it your page has html tags being display as well that means we need a more sophisticated text editor. I did a bit of digging and there is a editor in emailtemplate.html. And from your python you also need to have a restriction to only allow certain tags to prevent the text-editor from being hackable. image

@nahom70 this will be useful for you too

MImran2002 (changes requested):

app/templates/main/revisepositionpage.html
```diff
@@ -0,0 +1,106 @@
...
+                <p class="form-control-static">{{ position.revisedBy }}</p>
+              </div>
+            </div>
+          </div>
+
+          <h3 class="description-header">Description</h3>
+          <section id="sectionsContainer">
+            {%- for section in sections %}
+              <div class="description-section-row">
+                <div class="form-group">
```
brightfietsop-ux, line 63: Image 
So far, the font and the format are small issues that need to be addressed. There are lots of inconsistencies in it

app/templates/main/revisepositionpage.html
BhushanSah: This looks like the right direction. CKEditor is now loaded on the revise position page, and the JS initializes it for both existing sections and newly added sections. I also saw the backend sanitizer in `getPositions.py`, which is important because the editor alone does not make the submitted HTML safe.

Can we add a small test for `sanitizeDescriptionHTML()`? Since this content will be rendered as HTML later, I think we should make sure allowed tags are preserved and unsafe tags/attributes are stripped before this is merged.

BhushanSah (changes requested):

app/controllers/main_routes/departmentPortal.py
```diff
@@ -67,6 +67,43 @@ def downloadPositionDescription(org, account, positionCode):
 
 
 
+@main_bp.route('/department/<org>/<account>/positions/<positionCode>/revise', methods=['GET', 'POST'])
+def revisePosition(org, account, positionCode):
```
BhushanSah, line 71: Can we add an authorization check here before allowing the user to view or save a revision?

`managePositions` checks whether the user is a labor admin or a supervisor connected to this department, but this revise route does not appear to have the same check. Since this route can create a new position revision on POST, a user who knows the URL may be able to revise a position without going through Manage Positions.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -67,6 +67,43 @@ def downloadPositionDescription(org, account, positionCode):
...
+    try:
+        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+    except (NameError, DoesNotExist):
+        return render_template('errors/404.html'), 404
+
+    position = getPosition(dept, positionCode)
+
```
BhushanSah, line 78: I think this route should reuse the same department access logic as `managePositions`.

Something like this would keep the behavior consistent:

```python
if not g.currentUser.isLaborAdmin:
    if not SupervisorDepartment.select().where(
        (SupervisorDepartment.supervisor == g.currentUser.supervisor) &
        (SupervisorDepartment.department == dept.departmentID)
    ).exists():

File: app/logic/getPositions.py [hunk 13]
```diff
@@ -52,7 +54,56 @@ def getPositionDescriptionSections(position):
...
                                        .order_by(PositionDescriptionSection.order.asc()))
-    
+
     return positionDescriptionSections
 
+allowedDescriptionTags = {'p', 'br', 'strong', 'b', 'em', 'i', 'u', 'ul', 'ol', 'li', 'a', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'}
+tagPattern = re.compile(r'<(/?)\s*([a-zA-Z][a-zA-Z0-9]*)((?:\s+[^<>]*)?)\s*/?>')
+hrefPattern = re.compile(r'href\s*=\s*(["\'])(https?:.*?|mailto:.*?|/.*?)\1', re.IGNORECASE)
+
+def sanitizeDescriptionHTML(value):
```
BhushanSah, line 64: Can we add a small test for this sanitizer?

Since section content is edited with CKEditor and later rendered as HTML, it would be good to verify that allowed tags like `<p>`, `<ul>`, `<li>`, and `<strong>` stay, but unsafe tags/attributes like `<script>`, `onclick`, and unsafe links are removed.

app/logic/getPositions.py
```diff
@@ -52,7 +54,56 @@ def getPositionDescriptionSections(position):
     positionDescriptionSections = list(PositionDescriptionSection.select()
                                        .where(PositionDescriptionSection.position == position)
                                        .order_by(PositionDescriptionSection.order.asc()))
-    
+
     return positionDescriptionSections
 
+allowedDescriptionTags = {'p', 'br', 'strong', 'b', 'em', 'i', 'u', 'ul', 'ol', 'li', 'a', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'}
+tagPattern = re.compile(r'<(/?)\s*([a-zA-Z][a-zA-Z0-9]*)((?:\s+[^<>]*)?)\s*/?>')
```
BhushanSah, line 61: This sanitizer is a good start, but parsing HTML with regex can be fragile.

Could we either add stronger tests around tricky HTML cases, or use a dedicated HTML sanitizer if the project already has one available? The main thing I want to avoid is saving content with unsafe attributes or malformed tags that later get rendered with `|safe`.

app/controllers/main_routes/departmentPortal.py
```diff
@@ -67,6 +67,43 @@ def downloadPositionDescription(org, account, positionCode):
...
+        else:
+            position = createPositionRevision(
+                position,
+                g.currentUser.fullName,
+                request.form.get('positionTitle'),
+                wls,
+                request.form.getlist('sectionTitle[]'),
+                request.form.getlist('sectionContent[]')
+            )
+            flash('Position revision saved.', 'success')
```
BhushanSah, line 95: After saving a revision, should this redirect instead of rendering the page from the POST request?

Right now, refreshing the browser after a successful save could submit the POST again and create another requested revision. A redirect after save would avoid duplicate revisions from refresh.

app/logic/getPositions.py
```diff
@@ -52,7 +54,56 @@ def getPositionDescriptionSections(position):
...
+        positionTitle=positionTitle,
+        positionCode=position.positionCode,
+        department=position.department,
+        status="Requested",
+        wls=wls,
+        revisionDate=date.today(),
+        revisedBy=revisedBy
+    )
+
+    for order, (sectionTitle, sectionContent) in enumerate(zip(sectionTitles, sectionContents)):
```
BhushanSah, line 100: Can we validate that `sectionTitles` and `sectionContents` have the same length before using `zip()`?

If one list is longer than the other, `zip()` will silently drop the extra values. Since this is saving a revision, it would be safer to catch that case and return an error instead of losing part of the submitted form.

app/static/js/revisepositionpage.js
```diff
@@ -0,0 +1,32 @@
...
+        var row = fragment.querySelector('.description-section-row');
+        sectionsContainer.appendChild(fragment);
+        initEditor(row);
+    });
+
+    sectionsContainer.addEventListener('click', function (event) {
+        if (event.target.classList.contains('remove-section-btn')) {
+            var row = event.target.closest('.description-section-row');
+            var textarea = row.querySelector('textarea[name="sectionContent[]"]');
+            var editor = textarea && CKEDITOR.instances[textarea.id];
```
BhushanSah, line 25: Can we make sure this reliably destroys the CKEditor instance before removing a section?

The textareas in the template do not have explicit IDs, but this lookup depends on `textarea.id`. If CKEditor does not set the ID the way we expect, the editor instance may not be destroyed before the row is removed.

BhushanSah (changes requested):

app/templates/main/revisepositionpage.html
```diff
@@ -0,0 +1,106 @@
...
+
+            <div class="form-group row">
+              <label class="col-sm-3 col-form-label">Revised By:</label>
+              <div class="col-sm-9">
+                <p class="form-control-static">{{ position.revisedBy }}</p>
+              </div>
+            </div>
+          </div>
+
+          <h3 class="description-header">Description</h3>
```
MImran2002, line 59: Section Title still shows html tags 

Image

app/templates/main/revisepositionpage.html
```diff
@@ -0,0 +1,106 @@
...
+          </section>
+
+          <button type="button" id="addSectionBtn" class="btn btn-default">Add Section</button>
+
+          <div class="row revise-actions">
+            <div class="col-xs-12 text-left">
+              <a href="{{ url_for('main.managePositions', org=department.ORG, account=department.ACCOUNT) }}" class="btn btn-default">
+                Cancel
+              </a>
+              <button type="submit" class="btn btn-primary">Save Revision</button>
```
MImran2002, line 83: Save Revision doesn't portray the full picture save revision mean they can come back and edit. it should be submit revision.

app/templates/main/revisepositionpage.html
```diff
@@ -0,0 +1,106 @@
...
+            </div>
+          </div>
+
```
MImran2002, line 86: We didn't considered for the draft stage where our user instead of submit can save and leave only to come back. This means we need to add Draft as a status. This also mean we can use this draft and not change its composite key but its other fields. Moreover, in case the revisiondate move another day we can create a new one with different revision date base on the filter that the revision date is recent and draft exist if not we create a new one.

app/controllers/main_routes/departmentPortal.py
MImran2002, line 95: submitted for submission and saved for saving a draft

app/static/js/revisepositionpage.js
```diff
@@ -0,0 +1,32 @@
+$(document).ready(function () {
+    var sectionsContainer = document.getElementById('sectionsContainer');
+    var sectionRowTemplate = document.getElementById('sectionRowTemplate');
+
+    function initEditor(row) {
+        var textarea = row.querySelector('textarea[name="sectionContent[]"]');
+        if (textarea) {
+            CKEDITOR.replace(textarea);
```
MImran2002, line 8: if you use this here:             CKEDITOR.replace(textarea, {
                allowedContent:
                    'p br strong b em i u ul ol li a[href] h1 h2 h3 h4 h5 h6'
            }); you don't need to create a function, you don't need to sanitize the value in flask as this can sanitize

app/static/js/revisepositionpage.js
```diff
@@ -0,0 +1,32 @@
+$(document).ready(function () {
+    var sectionsContainer = document.getElementById('sectionsContainer');
+    var sectionRowTemplate = document.getElementById('sectionRowTemplate');
+
+    function initEditor(row) {
+        var textarea = row.querySelector('textarea[name="sectionContent[]"]');
+        if (textarea) {
```
MImran2002, line 7: the CKeditor that we add need other features for the use to add like list, bold, italic, font size and so on

app/static/js/revisepositionpage.js
```diff
@@ -0,0 +1,32 @@
+$(document).ready(function () {
+    var sectionsContainer = document.getElementById('sectionsContainer');
+    var sectionRowTemplate = document.getElementById('sectionRowTemplate');
+
+    function initEditor(row) {
```
MImran2002, line 5: we should also remove the insetplaceholder as we don't need it.

app/templates/main/revisepositionpage.html
```diff
@@ -0,0 +1,106 @@
...
+                <div class="form-group">
+                  <label>Section Content</label>
+                  <textarea class="form-control" name="sectionContent[]" rows="4">{{ section.sectionContent }}</textarea>
+                </div>
+                <button type="button" class="btn btn-danger btn-sm remove-section-btn">Remove Section</button>
+              </div>
+            {%- endfor %}
+          </section>
+
+          <button type="button" id="addSectionBtn" class="btn btn-default">Add Section</button>
```
MImran2002, line 76: this button should be green button

app/templates/main/revisepositionpage.html
```diff
@@ -0,0 +1,106 @@
...
+
+          <div class="row revise-actions">
+            <div class="col-xs-12 text-left">
+              <a href="{{ url_for('main.managePositions', org=department.ORG, account=department.ACCOUNT) }}" class="btn btn-default">
+                Cancel
```
MImran2002, line 81: cancel should be red and there should be a new button for submit which should be green too.

MImran2002 (changes requested):

PR #667, opened 2026-08-05
Create a page for departments to view their allocations

fritzj2:
## Issue Description

Fixes issue #607 
- Create a page for departments to view their current allocations. This should include both contracts and given allocations for the current term
image

## Additions

 - `allocationTable` `html`, `css`, and `js`, all handle the UI.
 - `allocationTable.js` is needed because there are 6 Bootstrap DataTables on the page.
 - `getTerms.py` has been added to handle getting the current term
 - `test_getTerms.py` to test the logic behind `getTerms.py`
## Changes

 - `main_routes.py` includes a section that navigates to the page
 - The `allocationManager.py` file includes 1 new function, `getContractedAllocations()`. This function gathers the number of hours for the break terms.
 - Lots of new objects in `demo_data.py,` which include more LSFs, formHistory, students, and terms.

## Rationale
 - This logic depends on having access to a term that aligns with the current year. `main_routes.py` contains a function that retrieves the current date and uses its year-end to determine the current term.
   - This term check flips on July 1st, as it is the day when the terms change for the labor office.
   - This check exists to get around there being no sense of a `currentTerm` in the system
 - The page will not physically load if there is no term for the current year.
 - The page also fails to load if there is no Allocation object for the current term.
 
 - Retrieving formHistory objects has a lot of nuance to it.
 - Contracts have many different ways that they can be written for only one term, including:
    - Changing term to Fall 2026 / Spring 2026
    - Changing the start / end date to the end of a semester
    - The kind of position (Primary vs Secondary)
    - The type of position (Weekly Hours vs Contract Hours)
  - The Reasoning behind the complex query is that there are many different ways in which a Labor Status form differentiates term contracts, yearly contracts, and break contracts.
  - The Supervisors can pick whichever way they choose to build the Labor Status Form, and the function needs to be prepared to deal with all of the cases present.

 - Breaks are hard to justify their place in the table. As nice as they are, it is hard to see the information that someone may want at a given time
 - The current implementation has all of the break allocations for a selected term. The problem is the summer term.
 - Summer term starts _after_ fall term.
   - The problem is that after July 1st, the next year's summer information will be displayed.
   - This is helpful as most of the longer summer term positions are finalized
     - As for the contracts that start after July 1st (think RAs who start ~August 1st), they can't be viewed since the term has rolled over.

- As for the `getTerms.py`, this does not grab break terms.
  - The reason that break terms are not included in this is that there are simply too many and too few use cases for them.
  - It is already hard to justify adding the spring and fall terms into the query; however, adding 5 more terms to it seems unnecessary in many cases and thus have been omitted.

## Testing
- Reset the database to test data (or the allocation grab function breaks the page)
- Navigate into the allocations page (home page -> Department Portal -> Computer Science -> Allocation Card -> View Allocations)
  - Other departments require an allocation model to be added for the page to load. 
    - `INSERT INTO allocation (id, department_id, isFinal, termCode_id, justification, primary_10, primary_12, primary_15, primary_20, secondary_5, secondary_10, breakHours) VALUES (222, 2, 1, 202600, "ETAD allocation", 1,1,1,1,1,1,111); will give you an allocation for the ETAD department`, which will load the page afterward.
- View the tables nested in the accordions, ensure that it is pulling `formHistory` and `allocations`
- Ensure that users not part of a specified department cannot access another department's page.
  - The quickest way to test this is by changing users in `secret_config.yaml`
- Mess with the data, ensure that the page loads with no contracts. 
- Check what happens with bad LSFs added to the data, such as with dates, terms, jobTypes, and contractHours.
- Check the test suite and ensure all tests are passing
- Check that the `test_allocationManager.py` test is passing, most importantly.


app/controllers/main_routes/main_routes.py
```diff
@@ -82,6 +85,71 @@ def departmentPortal(org=None,account=None):
                            positions = positionsList,
                            posURL = posURL)
 
+@main_bp.route('/department/<org>/<account>/allocations', methods=['GET'])
+def allocationTable(org=None, account=None):
+    currentUser = g.currentUser
+    try:
+        dept = Department.get(Department.ORG == org, Department.ACCOUNT == account)
+    except (NameError, DoesNotExist):
```
BhushanSah, line 93: Can we return a 404 here before continuing?

Right now, if the `org/account` does not match a department, `dept` stays `None`. Then the authorization block below can still try to use `dept.departmentID`, which would cause a server error instead of showing a normal 404 page.

app/controllers/main_routes/main_routes.py
```diff
@@ -82,6 +85,71 @@ def departmentPortal(org=None,account=None):
...
+        springTerm = Term.select().where(Term.termCode == currentDate.year * 100 + 12 - 100).get()
+        currentAY = Term.select().where(Term.termCode == currentDate.year * 100 - 100).get()
+        fallTerm = Term.select().where(Term.termCode == currentDate.year * 100 + 11 - 100).get()
+
+    else:
+        fallTerm = Term.select().where(Term.termCode == currentDate.year * 100 + 11).get()
+        currentAY = Term.select().where(Term.termCode == currentDate.year * 100).get()
+        springTerm = Term.select().where(Term.termCode == currentDate.year * 100 + 12).get()
+
+    allocationDict = getTotalAllocations(currentAY, dept)
```
BhushanSah: These helper functions look like they expect a `termCode` integer, but this route is passing full `Term` model objects.

Inside `allocationManager.py`, the helper calculates `academicYearCode` using `str(termCode)[:4]`, so passing a model object could break or depend on Peewee’s string representation. Can we pass the actual term codes instead?

```python
allocationDict = getTotalAllocations(currentAY.termCode, dept.departmentID)
fallContracts = getContractedAllocations(fallTerm.termCode, dept.departmentID)
springContracts = getContractedAllocations(springTerm.termCode, dept.departmentID)

File: app/controllers/main_routes/main_routes.py [hunk 3]
```diff
@@ -22,6 +24,7 @@
 from app.logic.banner import Banner
 from app.logic.getSupervisors import getSupervisors
 from app.logic.getPositions import getActivePositions
+from app.logic.allocationManager import *
```
BhushanSah: Can we avoid the wildcard import here?

It would be clearer to import only the functions used by this route, such as:

```python
from app.logic.allocationManager import getBreakContracts, getContractedAllocations, getTotalAllocations

File: app/controllers/main_routes/main_routes.py [hunk 4]
```diff
@@ -82,6 +85,71 @@ def departmentPortal(org=None,account=None):
...
+                           department = dept,
+                           currentAY = currentAY,
+                           allocations = allocationDict,
+                           fallContracts = fallContracts,
+                           springContracts = springContracts,
+                           breakContracts = breakContracts)
+                           
+
+@main_bp.route('/supervisorPortal/addUserToDept', methods=['GET', 'POST'])
+def addUserToDept():
```
BhushanSah: This route seems unrelated to the allocation table feature.

Can this be removed from this PR or moved to the PR that needs the supervisor/member add behavior? Keeping unrelated route changes out of this PR would make the allocation work easier to review.

app/logic/allocationManager.py
```diff
@@ -44,6 +44,30 @@ def countContracts(jobType: str, weeklyContractHours: int, termCode: int, dept:
     5-hour positions in the CS department for the 2025 Fall term.
     '''
     academicYearCode = int(str(termCode)[:4] + "00") 
+
+    # This sets the date condition to determine whether the form is within the boundaries of the term
+    # Fall only contracts end before spring, spring contracts start after fall.
+    fallMonths = ["07","08","09","10","11","12"]
```
BhushanSah, line 50: Can we clean up this date filtering?

This currently compares extracted month values to string values like `"07"` and `"08"`. It may be safer to compare against actual dates or the selected term’s `termStart` / `termEnd` instead of only checking months. Month-only logic can get confusing for contracts that cross years, cover the full academic year, or overlap summer.

app/logic/allocationManager.py
```diff
@@ -105,4 +130,18 @@ def getContractedAllocations(termCode: int, dept: int):
...
+
+def getBreakContracts(termCode, dept):
+    break_allocation = FormHistory.select(fn.SUM(LaborStatusForm.contractHours)
+    ).join(LaborStatusForm
+    ).where(
+        FormHistory.historyType == "Labor Status Form",
+        FormHistory.status.in_(["Approved", "Pending", "Pre-Student Approval"]),
+        LaborStatusForm.termCode == termCode,
+        LaborStatusForm.department == dept,
+        LaborStatusForm.contractHours != None).scalar()
```
BhushanSah: Small Peewee cleanup: can we use `.is_null(False)` here instead of `!= None`?

```python
LaborStatusForm.contractHours.is_null(False)

File: app/templates/main/allocationTable.html [hunk 7]
```diff
@@ -0,0 +1,257 @@
...
+    <script rel="stylesheet" type="text/css" href="/static/css/allocationTable.css?u={{lastStaticUpdate}}"></script>
+{% endblock %}
+
+{% block app_content %}
+
+<h2 class="text-center">{% if department %} {{department.DEPT_NAME}} Allocations {% else %} Choose a Department: {% endif %} </h2>
+
+<div class ="button-container">
+  <h3 class="termDisplay">Current Term: {{currentAY.termName}}</h3>
+  <a id="download" class="btn btn-success" href="/department/{{department.ORG}}/{{department.ACCOUNT}}/allocations/download">Download Allocation History</a>
```
BhushanSah, line 21: I do not see routes for `/allocations/download` or `/allocations/request` in this PR.

Can we either add those routes or remove/disable these buttons for now? Otherwise users can click them and hit a 404.

tests/code/test_tracy.py
```diff
@@ -18,8 +18,11 @@ def test_init(self, tracy):
     def test_getStudents(self, tracy):
         with app.app_context():
             students = tracy.getStudents()
-            assert ['Elaheh','Guillermo','Jeremiah','Kat', 'Oluwagbayi', 'Test', 'Tyler'] == [s.FIRST_NAME for s in students]
-            assert ['718','300','420','420', '883', '700', '420'] == [s.STU_CPO for s in students]
+            for student in ['Elaheh','Guillermo','Jeremiah','Kat', 'Oluwagbayi', 'Test', 'Tyler']:
+                assert student in [s.FIRST_NAME for s in students]
```
BhushanSah, line 22: This makes the Tracy test less strict.

The old test checked the exact returned names and CPOs. The new version only checks that each expected value appears somewhere, so it would still pass if extra students are returned or if names and CPOs no longer match the same records.

Can we keep this test stricter, or update it in a way that still verifies the full expected result?

tests/code/test_allocationManager.py
```diff
@@ -48,6 +48,15 @@ def testTerm():
     #destroy
     term.delete_instance()
```
BhushanSah, line 49: Small naming issue: this file is named `test_allocationManger.py`. Can we rename it to `test_allocationManager.py` so it matches `allocationManager.py`?

BhushanSah (changes requested):

app/templates/main/allocationTable.html
fritzj2: This is not in the scope of this issue; check out  #668 for the branch that adds the route for the allocation request.

app/logic/allocationManager.py
fritzj2: As interesting a concept as it sounds, the impact on code performance here is almost negligible. In the end, the check for the year-long positions checks the end-of-year dates anyway. e.g. (lsf.startDate >= AY.startDate) & lsf.startDate <= AY.endDate).
 - In the end, it requires a comparison that changes the implementation to check the same thing. It still checks if it pushes into the end of the year regardless.
 - To change this implementation to be more strict, it requires us to always pass in a fall and spring term into the code, which we may not always need. This implementation is also the longest of all the options and the hardest to read.
 - The edge case that this fixes, however, is a termDate change. Where the end of a Fall term ends in January, instead of December.

The end code requires many comparisons, which makes the already hard-to-read code even harder to read and debug, for very little benefit.

tests/code/test_tracy.py
fritzj2: ### For the Semester Team

This change was originally meant to fix an extremely brittle test, which, thanks to my `demo_data.py` changes, now fails
There are a few ways to go around this, such as adding atomic students to the test rather than relying on existing database students, or checking a small subset of the data instead of the entire database.

This is the function that adds students to Tracy in `demo_data.py`; this could be used to make atomic students.
`# Add students to Tracy db`
`with app.app_context():`
   ` for student in (tracyStudents + bothStudents):`
    `    db.session.add(STUDATA(**student))`
    `    db.session.commit()`

`# Add the Student records`
`students = []`
`for student in (localStudents + bothStudents):`
`    # Set up lsf db data`
`    del student["PIDM"]`
`    student['ID'] = student['ID'].strip()`
`    student['legal_name'] = student['FIRST_NAME'].strip()`
 `   del student['FIRST_NAME']`

`    students.append(student)`
`Student.insert_many(students).on_conflict_replace().execute()`

app/controllers/main_routes/main_routes.py
```diff
@@ -1,6 +1,7 @@
 from flask import render_template, request, json, redirect, url_for, send_file, g, flash, jsonify
 from peewee import JOIN, DoesNotExist, fn
```
MImran2002: Image the data tables contracted after a bit when the drop down is open. Try looking into why that is happening.

app/controllers/main_routes/main_routes.py
```diff
@@ -1,6 +1,7 @@
...
 from functools import reduce
+from datetime import datetime, date
```
MImran2002: Total Contracts row should have a tooltip or event like Primary + Secondary total Contract to show that total contracts mean a combination of both as here you have for both primary and secondary table and that can confuse anyone who are not familiar with the term contract as this isn't labor who is using it is all the departments and supervisor who might not be familiar with the term.

Image

app/controllers/main_routes/main_routes.py
```diff
@@ -22,6 +24,7 @@
 from app.logic.banner import Banner
 from app.logic.getSupervisors import getSupervisors
 from app.logic.getPositions import getActivePositions
+from app.logic.allocationManager import getBreakContracts, getContractedAllocations, getTotalAllocations
```
MImran2002: have your accordion words to the left 

Image

app/controllers/main_routes/main_routes.py
```diff
@@ -82,6 +85,55 @@ def departmentPortal(org=None,account=None):
...
+        
+    if not currentUser.isLaborAdmin:
+        allowedDepartmentIds = [d.departmentID for d in getDepartmentsForSupervisor(currentUser)]
+        if dept.departmentID not in allowedDepartmentIds:
+            return render_template('errors/403.html'), 403
+
+
+    currentDate = date.today()
+    if currentDate.month <= 6:
+        # If it is the spring semester, then the term code is 1 year behind. e.g. 2025-2026 term code is 202500. Thus the - 100 in the spring term.
```
MImran2002: can be in a logic file and have test, I am not going to require you as this is simple logic but many simple logic. However having different combination of math after currentDate.year means this can be turn into a logic function that determines the termcode and we will see similar termcode logic being use in other branches like allocation request.

MImran2002 (changes requested):

MImran2002:
Aside from the requested these are the scenarios I have checked and what i can say is it works, different term forms don't affect the current ay term display. when a lsf is submitted the display shows corrected and when i modify the form into different hours the display shows correctly too.

app/controllers/main_routes/main_routes.py
fritzj2: The tables no longer resize, this was an issue caused by css having padding that would not update until the container was fully opened.

app/controllers/main_routes/main_routes.py
fritzj2: Added a tool-tip to each category of the Total contracts which reads "Total Contracts shows the count for both Primaries and Secondaries for the term". 
It may also be worthwhile to consider removing the Total Contracts from the secondaries table to avoid the redundancy.

app/controllers/main_routes/main_routes.py
fritzj2: I have made `getTerm.py` to handle the logic behind this; it does more than just get the current term. Tossing in an academic year gives the user the AY, fall, and spring terms for that Academic year. If the Academic year is not provided, it gets the current term.

MImran2002 (approved):

